PR #134, opened 2014-12-20
Jet Validation Studies tools

ldcorpe:
Hi Seth,

As agreed, here is a PR for the new JetValidation studies folder, along with the various tools and configs we have created. The folder has been cleared up from local versions to have just one config which has all the jet energy corrections applied. The tree maker in the plugins folder creates everything needed for comparisons between the 5 jet collections, as well as efficiency and fake rate plots. The information required for photon-dR veto is included, as is information for a PUJetID cut. There is also a folder for the various plotting macros which we have been using to generate plots.

All the best and merry Xmas,

Louie


sethzenz:
Replaced by #135

PR #140, opened 2015-01-08
Additions to PhotonIdUtils/PhotonProducer:

musella:
- recomputation of non-ZS (aka full 5x5) cluster shapes (needed for Phys14 miniAOD)
- optionally also use non-ZS for extra flashgg::Photon cluster shapes
- implemented MC match classification


sethzenz:
Merges successfully, compiles, MicroAOD and Tag tests run.

PR #141, opened 2015-01-08
implementation of VHhadronicTag format and producer

favaro:


sethzenz:
Replaced by #142, whose content is identical

PR #142, opened 2015-01-08
added VHhadronicTag to TagFormats and TagProducers

favaro:


sethzenz:
Identical to #141, compiles, MicroAOD and Tag tests run, and new tag is created and read successfully.

PR #144, opened 2015-01-09
Added extra MC truth

musella:
- PhotonMCUtils class to compute gen-level isolation and truth-determination algos
- Added GenPhotonExtra collection to miniAOD to store gen-level info
- Optionially to copy extra info to flashgg::Photon
- Factorized gen-related info into dedicated sequence

Also made PhotonProducer parameters tracked such that settings are stored in provenence info
by CMSSW.

Note: having the GenPhotonExtra in the event we should be in a position to drop the packedGenCandidates, except for jet flavour tagging (which will be eventually be needed for b-tagging systematic uncertainties).

Apologies, I for not having included this in #140. I though it would have taken me longer to complete this stuff, so I went ahead with #140.


sethzenz:
I think you have forgotten to include MicroAODFormats/interface/GenPhotonExtra.h in this PR?

musella:
Indeed I had forgotten to add the new files to the repository.

They are there now. Sorry about that.

sethzenz:
Merges successfully, compiles, MicroAOD and Tag tests run.

Issue #111, opened 2014-12-01
Tags: interleaved sorting


sethzenz:
Allow, for example, VBF 0 > Untagged 0 > VBF 1 > ...    Configure TagSorter with VPSet list of tuple(tagName,minCat,maxCat).

sethzenz:
Addressed by number #146

PR #146, opened 2015-01-16
modify TagSorter to allow priority to depend on category number

sethzenz:
Add TagSorter option to "interleave" tags based on category number, e.g. VBF 0 > Untagged 0 > Untagged 1 > VBF 1 > VBF 2 > Untagged 3 > ...   Python interface is changed.

flashggTagSorter_cfi.py is modified to illustrate the new interface, including an example of "interleaved" categories.


sethzenz:
Addresses #111

PR #150, opened 2015-01-23
72 xnew

yhaddad:
Setting up the MicroAODProducer for the Grid production. This pull request includes : 
1- the configuration files for the Crab3
2- Fix memory issues related particularly to PFCH0 Producer
3- Increase the memory usage on the Grid
4- New configuration file for the JetValidationTree producers


sethzenz:
This needs a few changes to make sure it works with the existing recipes and organization.  Once the changes are made, you can simply commit again to the same branch and push to your repo; when you do, this pull request will automatically "pick up" the changes.   Here's what is needed:

1. remove (using git rm) this file: MetaData/work/HggPhys14_jets/simple_Producer_jets.py -- this is the wrong place for a python config, and in any case you have an identical copy elsewhere

2. move (using git mv) the following files into JetValidation/configs:

MicroAODProducers/test/simple_Producer_jets.py
MicroAODProducers/test/simple_Producer_jets_local.py

3. revert your change to this file, which added a "keep" line for jets that aren't usually produced:

MicroAODProducers/python/flashggMicroAODOutputCommands_cff.py

Instead of adding the keep line to the default, it should be added where needed to the individual configurations that rely on it, as an addition to the output commands AFTER the default output commands are added.

4. change back the parts of MicroAODProducers/plugins/JetProducer.cc that are related to the experiments with the modified version of Pileup Jet ID; specifically:

* Loading of PF Candidate collections (make sure to remove applicable lines in all parts of code)
* Passing of PF Candidate collections to pileup jet id

sethzenz:
Tres bien!  Compiles, usual MicroAOD and Tag tests pass.

PR #154, opened 2015-02-02
tth leptonic producer; small bug fixing in electrons

InnaKucher:


sethzenz:
compiles, simple producer test runs, simple tag tests run, tth leptonic tag seen. looks good.

Issue #153, opened 2015-01-30
Add electron ID and MVA cuts to tags using electrons


sethzenz:
Since PR #152 there are no longer any cuts to electrons in the ElectronProducer.  (I kept the code but added an ApplyCuts flag that is set to false by default.)  These cuts should be put into the Tag Producers that use electrons instead.

sethzenz:
TTHLeptonicTagProducer, VHLooseTagProducer, and VHTightTagProducer all appear to do this in the current version of the code.

PR #155, opened 2015-02-02
VH producers; adding electrons

InnaKucher:
I tested both with MicroAOD_plus_Tag_test.py config file.


sethzenz:
Hi @InnaKucher, I seem to have found an issue.

I merged this PR my local copy of the current upstream/master.  I ran simple_Producer_test.py and simple_Tag_test.py separately.  Many events were ok but I then got this complaint:

%MSG-e FatalSystemSignal:  FlashggVHtightTagProducer:flashggVHtightTag  04-Feb-2015 12:07:35 CET Run: 1 Event: 5349
A fatal system signal has occurred: segmentation violation
%MSG

I assume this showed up in my tests but not yours because the input samples are different...

My input: /store/mc/Phys14DR/GluGluToHToGG_M-125_13TeV-powheg-pythia6/MINIAODSIM/PU20bx25_tsg_PHYS14_25_V1-v1/00000/3C2EFAB1-B16F-E411-AB34-7845C4FC39FB.root

Combined test input: /store/mc/Spring14miniaod/WH_ZH_HToGG_M-125_13TeV_pythia6/MINIAODSIM/PU20bx25_POSTLS170_V5-v2/30000/E0D066C6-2219-E411-BD9A-02163E00ECDF.root

Can you switch to the input file I used and try to reproduce the seg fault and find its origin? Thanks!

InnaKucher:
I used those miniAOD:
https://cmsweb.cern.ch/das/request?input=dataset%3D%2FWH_ZH_HToGG_M-125_13TeV_pythia6%2FPhys14DR-PU20bx25_PHYS14_25_V1-v1%2FMINIAODSIM&instance=prod%2Fphys03

I'll check with yours.

InnaKucher:
It seems ok for me. I'll check what else I have to commit!

sethzenz:
Hmm, yes, I get the seg fault even when I switch inputs and producer files.  My area has no difference from InnaKucher/master (using git diff) so maybe you did miss committing some change.

InnaKucher:
I know the problem. If you put this to "true" everything works fine.

https://github.com/cms-analysis/flashgg/blob/master/MicroAODProducers/plugins/ElectronProducer.cc#L60

 I tested with your input 

/store/mc/Phys14DR/GluGluToHToGG_M-125_13TeV-powheg-pythia6/MINIAODSIM/PU20bx25_tsg_PHYS14_25_V1-v1/00000/3C2EFAB1-B16F-E411-AB34-7845C4FC39FB.root

sethzenz:
OK, thanks for tracking that down.  However, by default we want that flag set to false -- that is, we want to leave the exact electron selection to the downstream tag producers, per issue #153.  (I am happy to discuss this, but running tight cuts in the flashgg ElectronProducer would be a different approach than is used for all the object producers and could potentially interfere with validation, which is why I put that flag in.)

sethzenz:
So that means that the Tag producers that use electrons, including this one, should apply the relevant cuts themselves -- and avoid crashing, of course!

InnaKucher:
I don't understand. If we keep Electron Producer, it means we apply those cuts in Electron Producer. 
I thought you put this flag "applyCuts_" only for electron validation. 

So I have to keep applyCuts = false and cut on mvaID and iso in my producer? We will keep like this?

InnaKucher:
I see. I will apply those cuts in producers and commit again. Thank you for explaining me this.

sethzenz:
Sorry if I was unclear about this when we discussed it before.  Here's the situation as I see it:

1. None of the other producers apply cuts; they only add relevant information.

2. In the future we might add some preselection cuts to save space, but the plan is not currently to have the final analysis cuts applied in any object producer. 

3. Even if we changed the final plan, Tag producers cannot rely on the cuts being applied in order to run correctly, because we will often need to loosen the cuts in order to test various parts of the code.

4. Therefore each Tag object should be interpreting the event using the cuts that are appropriate to its own analysis category.

sethzenz:
Excellent, thank you.  If you commit to your master branch then the changes should be added to this PR automatically.

InnaKucher:
Thank you, Seth! I just didn't see your comment with explanation. After I saw ans understood. Sorry.

InnaKucher:
Eta cut also should be implemented in taggers? 

https://github.com/cms-analysis/flashgg/blob/master/MicroAODProducers/plugins/ElectronProducer.cc#L128

sethzenz:
I left it in the Electron Producer, because Aeff and isolation are not defined outside the crack or beyond the tracker. I can't imagine we'd ever want to loosen that cut.  But it's probably wise to have it in the tagger too, for completeness.

InnaKucher:
I can keep Aeff definition, depending on eta in Electron Producer as it is:
https://github.com/cms-analysis/flashgg/blob/master/MicroAODProducers/plugins/ElectronProducer.cc#L130-L136

I can remove just this eta cut : https://github.com/cms-analysis/flashgg/blob/master/MicroAODProducers/plugins/ElectronProducer.cc#L128
and I do this cut in my tagger. 
Or it doesn't matter and I can keep this cut in producer (I mean we do eta cuts are because of gaps in the detector, they are not going to change) ?

sethzenz:
The eta cuts are based on the detector so I expect them not too change.  But I think it is OK to put it in either place.

InnaKucher:
Another question about this one: 
https://github.com/cms-analysis/flashgg/blob/master/MicroAODProducers/plugins/ElectronProducer.cc#L147

Can I save it to flashgg::Electron also. If not, I have to put this stuff to each producer :

https://github.com/cms-analysis/flashgg/blob/master/MicroAODProducers/plugins/ElectronProducer.cc#L101-L109

What do you think?

sethzenz:
Yes, I think it is a very good idea to save a hasMatchedConversions variable to the Electron itself.

InnaKucher:
Ok, I will do like this: save hasMatchedConversions to Electron and put cut on eta outside the Electron Producer.

InnaKucher:
Done. Tested with 2 files your wrote in the first comment. Tag producers don't crash on them.

sethzenz:
This merges and works.  Thanks for going back and forth on all this!

PR #156, opened 2015-02-05
Final fit folders

ldcorpe:
Hi Seth,

Attached is a pull request for the FinalFit directories, which I have adapted from h2gglobe to compile as standlaone in the FLASHgg environment. Four new folders:
*FinalFitSignal: Renamed from SimultaneousSignalFit as discussed. Major changes include modification of makefile to compile in CMSSW7xx/flashgg environment (had to add a few extra libraries, and remove dependencies like loopAll). Also modified to not block FLAShgg compilation when using scram b: had to remove ClassImp statements for new RooFit class and make BuildFile.xml to load correct libraries.
*FinalFitBackground: Renamed from BackgroundProfileFitting. Major changes include small modifs to compile as standalone with make and not block scram b compilation of FLASHgg. This included initialising uninitialised  variables (marked with comments where I did this) and removing ClassDef and ClassImp statements (This appears not to hinder make compilation of executables...), and creating a BuildFile.xml to load relevant libraries.
*FinalFitDatacard: Moved relevant macros from h2gglobe/Macros pertaining to datacard creation.
*FinalFitPlots: Renamed from h22globe/Macros. Initially I kept the bulk of the contents, but in the end I removed the vast majority of the macros by semi-randomly judging whether they could be re-used. Any macros which I have removed but which are indeed needed can be obtained from : https://github.com/ldcorpe/h2gglobe/tree/master/Macros

All the best,

Louie


sethzenz:
N.B. These additions have no impact on any existing flashgg code.  However please note that README.md is updated: the CombinedLimit package must be obtained, or compilation will fail.

PR #157, opened 2015-02-09
Added (di-)photon dumpers for full framework and  generic cut-based di-photon selector.

musella:


sethzenz:
Does not compile for me.  I get messages like this:

`undefined reference to 'reco::parser::cutParser(edm::TypeWithDict const&, std::string const&, boost::shared_ptr<reco::parser::SelectorBase>&, bool)'`

Is there code missing?  I checked for a change in instructions requiring a release change or new topic branch, but maybe I missed it...

musella:
Sorry, forgot Buildfile changes...

sethzenz:
OK, builds now, thanks!  Will merge now.

PR #160, opened 2015-02-11
Import few tools for Phys14 studies

musella:
- Added global variables in CollectionDumper
- Single photon dumper and configuration for photonID ntuples
- Some improvements to MetaData system


sethzenz:
I think there is a file missing from this PR? TagAlgos/interface/GlobalVariablesDumper.h

PR #164, opened 2015-02-12
Fix the charged  pf iso calculation 

JunquanTao:
pfchargediso wo e/mu, commit id ff89d5d3f8a525168ed54625f65532e3d59bd152


sethzenz:
This PR has some extra stuff due to divergence and merge.   @JunquanTao and I will discuss fixing it offline.

PR #173, opened 2015-02-20
Topic few fixes for Zee

matteosan1:
Added typedef for CutBasedPhotonDumper.
Added few missing accessors: ID MVA in diphoton and chosenvtx isolation cone 0.3 in photon.


sethzenz:
Looks good, passes usual tests

PR #179, opened 2015-03-02
conversion-tools-for-pat-ele-setup

InnaKucher:
Setup file with new method for pat electrons

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -81,5 +81,10 @@ git cms-addpkg RecoEcal/EgammaCoreTools
 git cherry-pick fd6b6bd88dc74d8f73299f395cc075aa89c8d701 
 
 echo
+echo "Setting up Conversion Tools for pat ele..."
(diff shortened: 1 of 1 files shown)


sethzenz:
The addition to the script builds ok, so this is fine.  However, no flashgg code uses the change yet.  So we will keep this request on hold until there's a PR for commits that need it.

sethzenz:
Also just to note: at some point a text editor (in Windows?) touched this file that added a '\r' character at the end of every line, causing the diffs to claim every line changed even though only a few lines were added.   Best to avoid this if possible; it can be checked by diffing and seeing if the whole file is changed.

InnaKucher:
It is not Windows, I use Linux and usually edit files with vim. Ok, I will check files in future, thanks.

sethzenz:
See #180 for discussion of how to fix the extra character and preserve history

PR #180, opened 2015-03-03
App ele veto

InnaKucher:
I can add the same line to all Tag Producers as for TTH hadronic.

### DIFF
--- a/MicroAODFormats/interface/Muon.h
+++ b/MicroAODFormats/interface/Muon.h
@@ -1,0 +1,31 @@
+#ifndef FLASHgg_Muon_h
+#define FLASHgg_Muon_h
+
+#include "DataFormats/PatCandidates/interface/Muon.h"
+#include "flashgg/MicroAODFormats/interface/DiPhotonCandidate.h"
+#include "DataFormats/VertexReco/interface/Vertex.h"
+
+namespace flashgg {
+
+	class Muon : public pat::Muon {
+
+		public:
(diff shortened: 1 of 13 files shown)


sethzenz:
Hi @InnaKucher, this PR has the same problem as the last one: several files have an invisible extra character added, so that the diffs appear to show the whole file being deleted and replaced.  This will make it hard to use the history in the future.  Is it possible to fix this and do a new PR from scratch?

There are two things that ideally need to be done:
1. Remove '\r' from everywhere it appears in every file.  I can provide a simple python script to do this if you would like one.
2. Produce a new branch and rebase so that the history does not show all the lines changing and changing back, but rather shows only the lines that have actually been changed.

Sorry this glitch is making extra work for you!

InnaKucher:
Hi, yes python script can be very useful. They you. I forgot to check files with diff.

sethzenz:
OK, you can use the script here:

https://sethzenz.web.cern.ch/sethzenz/cms/remove_r.py

Usage:

`python remove_r.py [input] [output]`

For example:

`python remove_r.py setup.sh setup_fixed.sh`

You'll then have to copy the output back over to the right filename if it looks ok.  This is because I didn't want the script itself to have the possibility of screwing anything up. 

I will close this PR and the other one, because you'll have to create a new branch and rebase in order to keep the history from containing both the appearance and disappearance of the extra character.

PR #185, opened 2015-03-09
Switch Jet Collection to PFCHSLeg as default

ldcorpe:
Hi,

This PR switches the default Input Jet Collection for FLASHggJets from PFCHS0 (the miniAOD default, pre-clustered Jets which do Charged Hadron Subtraction using the 0th Vertex in the PV collection) to PFCHSLeg ( now re-clustering Jets, following the CHS procedure but with "legacy" (ie Hgg-selected) vertex rather than 0th Vertex). This now is consistent with the current PFCHS0 JEC, but it also logically conssitent with the vertex we use for the rest of the analysis.

Any issues, please shout,

Louie

### DIFF
--- a/JetValidation/configs/simple_Producer_jets.py
+++ b/JetValidation/configs/simple_Producer_jets.py
@@ -149,7 +149,7 @@ process.combinedSecondaryVertex.trackMultiplicityMin = 1  #needed for CMSSW < 71
 process.flashggCHSLegacyVertexCandidates = cms.EDProducer('FlashggCHSLegacyVertexCandidateProducer',
                                                           PFCandidatesTag=cms.untracked.InputTag('packedPFCandidates'),
                                                           DiPhotonTag=cms.untracked.InputTag('flashggDiPhotons'),
-                                                          VertexCandidateMapTag = cms.InputTag("flashggVertexMapUnique"),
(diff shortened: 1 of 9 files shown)


sethzenz:
I think we can refactor the add jet collections function a bit more later, but it's mostly ok for now.  However one bit needs to be looked at and fixed:

https://github.com/ldcorpe/flashgg/blob/5032c5328183a4c493ae1f779d659dd209854930/MicroAODProducers/python/flashggJets_cfi.py#L10-L15

1. Check for anything being added there that's already added at the top of the top-level config, and remove it from the jet function.
2. Do not change the global tag (that's definitely already in the top-level config)

ldcorpe:
Ok, I have dealt with your suggestion and trimmed down the files being loaded to bare minimum. Any further issues let me know!

sethzenz:
Tests are fine.  Please note that we now have this (100% accurate) warning from time to time:

`[WARNING] Multiple Diphotons in event, with different PVs . Using 0th Diphoton Vtx for CHS PU subtraction.`

Once diphoton preselection is used, these will be much rarer, but we may still need to be smarter about having multiple collections if we end up using PFCHS in the analysis.  In any case, even the default behavior in the case of this warning is better than the old PFCHS0.

PR #186, opened 2015-03-09
fix electron veto with pat electron

InnaKucher:


InnaKucher:
Electron veto for pat electrons. 
\r issue fixed.

PR #191, opened 2015-03-12
Refurbishing

ferriff:
- redistribute files according to [1]
- change some class (file)names for consistency with the flashgg notation
- modify the object methods from getSomething() into something()
   . only drawback: DiPhotonCandidate::getVertex() becomes DiPhotonCandidate::vtx() as the vertex() method is already inherited from reco::Candidate (as well as an additional vertex data member)

N.B. has to be sync'ed with the update to Dummy/Puppi to account for corresponding changes in flashgg

[1] mostly https://cern.ch/go/9Chd and related discussions


sethzenz:
This replaces #190, correct?

ferriff:
Yes, #190 was open by a wrong button push that I see only now.
Federico

Issue #163, opened 2015-02-12
Migrate to 73X


sethzenz:
It appears that CMSSW_7_3_2 is a sufficiently stable migration target.  However, some of the effort (e.g. on jet tools) is non-trivial.  I propose we do this after the Higgs Workshop to avoid confusion.

sethzenz:
We ended up going to CMSSW_7_4_0_pre9

musella:
Hi,

sorry to reply after this has been closed, but I think that passing
edm::Handles around in interfaces is not a good practice, especially if
they are not const.

This makes the code tight to the fact that objects are read/written in the
event, while this may not necessarily be the case.

I would rather suggest that PtrVectors are replaced with const vector<Ptr>
in interfaces.
At most, one should pass direcly Views around.

Cheers,
Pasquale





On 26 March 2015 at 16:34, Seth Zenz <notifications@github.com> wrote:

> Closed #163 <https://github.com/cms-analysis/flashgg/issues/163>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cms-analysis/flashgg/issues/163#event-265991405>.
>



-- 
Pasquale Musella - ETHZ
http://musella.web.cern.ch/musella

sethzenz:
Yes, I agree with this.  At first we didn't realize that vector<Ptr> was an option even though PtrVector no longer is, but we should still go back and keep the Handles out of the interfaces.  I've created #202 for this issue.

Issue #57, opened 2014-10-04
Change object format for DzVertexMap object


sethzenz:
TLDR: we should get rid of maps we put in the event.  I will take care of this but let me know if you need a map and I can help you replace it.  More details below if you're curious.

I have been advised by CMSSW memory management experts that maps can create issues or be unstable, and that `map<T,vector<U> >` is especially bad. [Here T=`Ptr<vertex>` and U=`PackedCandidate`.]  The use of this construct for a one-to-many map can be replaced with some clever handling of `vector<pair<T,U> >`, in which all the U's corresponding to a given T are in sequence and you rely on that to look at only the right part of the vector.

Also, for reference, a one-to-one `map<T,U>` should simply be replaced with a vector<U> for which you rely on the fact that the new collection is in the same order as the old. I had significant trouble creating the map anyway, because CMSSW implicitly requires a `vector<T>` dictionary for a `map<T,U>` even if you don't think you're using `vector<T>`.  I eventually got rid of the map (in code I will do a pull request for later this morning) and just used the `vector<U>`.

sethzenz:
std::unordered_multimap<T, U> now seems to me to be a more intuitive choice.

PR #201, opened 2015-03-26
Topic flashgg 74x porting

ldcorpe:
Summary of changes

*Made PFCHSLeg default collection, using updated AddJetCollection tool.

*Removed PUJetID and set all jets to pass temporarily because of buggy code from JME.

*Removed references to edm::PtrVector as this is now obsolete. Now loop using edm::Handle<edm::View< object> > directly and access items using ptrAt(index) method. (Previously used ptrVector() method to get a PtrVector, the looped over using PtrVector[index]). 

*In other cases where this is not possible to simply use edm::Handle < edm::View < > >, use std::vector<edm::Ptr< object > >.

*Updated dictionaries correspondingly.

*Vertex-Candidate map reworked by Seth.

Thanks,

Louie


sethzenz:
Addresses #57 and #163

sethzenz:
OK, this is going in, and everything should now compile and run self-consistently.  I'll move the instructions to this version also.  (A new PR will come from Louie later restoring the pileup jet id.)

musella:
Hi,

sorry to reply after this has been closed, but I think that passing
edm::Handles around in interfaces is not a good practice, especially if
they are not const.

This makes the code tight to the fact that objects are read/written in the
event, while this may not necessarily be the case.

I would rather suggest that PtrVectors are replaced with const vector<Ptr>
in interfaces.
At most, one should pass direcly Views around.

Cheers,
Pasquale


On 26 March 2015 at 16:33, Seth Zenz <notifications@github.com> wrote:

> Merged #201 <https://github.com/cms-analysis/flashgg/pull/201>.
>
> —
> <https://github.com/cms-analysis/flashgg/pull/201#event-265990150>.
>



-- 
Pasquale Musella - ETHZ

PR #205, opened 2015-04-01
get sigEOverE from regression in reco (default energy already regressed)

sethzenz:
Regression and regression errors are defined in reco and the code is now setup to use them.  Note that the following are equal by definition (for fg a `flashgg::Photon`):

`fg.energy() == fg.getCorrectedEnergy(flashgg::Photon::regression_type)` [set in reco]
`fg.sigEOverE() == fg.getCorrectedEnergyError(flashgg::Photon::regression_type)/fg.energy()` [flashgg]

### DIFF
--- a/DataFormats/interface/Photon.h
+++ b/DataFormats/interface/Photon.h
@@ -49,7 +49,6 @@ namespace flashgg {
     void setpfChgIsoWrtChosenVtx03(float val) {pfChgIsoWrtChosenVtx03_=val;};
     void setESEffSigmaRR(float val) {ESEffSigmaRR_=val;};  
     void setPhoIdMvaD( std::map<edm::Ptr<reco::Vertex>,float> valmap ) {  phoIdMvaD_ = valmap; };   // concept: pass the pre-computed map when calling this in the producer
-    void setEnergyAtStep(std::string key,float val);
     void updateEnergy(std::string key,float val);
     //    void setSigEOverE(float val) { sigEOverE_ = val; };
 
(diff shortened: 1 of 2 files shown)


DataFormats/interface/Photon.h
```diff
@@ -85,7 +88,7 @@ namespace flashgg {
 
     bool hasEnergyAtStep(std::string key) const;
     float const energyAtStep(std::string key) const;
-    float const sigEOverE() const { return sigEOverE_; };
+    float const sigEOverE() const { return (getCorrectedEnergyError(regression_type) / energy()); } // Regression + error now in reco and base object
```
musella: Isn't there a sigmaEoverE variable already stored in reco?

Computing sigmaEoverE as sigmaE/E has the problem that the denominator changes upon application of energy scale corrections/smearing.

If the relative resolution is not available, I would repalce energy() with the appropriate energyAtStep() call.

sethzenz:
This is a good point.  In reco it appears the error itself is stored:

https://github.com/cms-sw/cmssw/blob/CMSSW_7_5_X/DataFormats/EgammaCandidates/src/Photon.cc#L144-L161

So I will use energyAtStep("initial") as you suggest.

sethzenz:
Actually I'll do: `getCorrectedEnergyError(regression_type) / getCorrectedEnergy(regression_type)` 

Those values are stored separately from the current P4 energy, and later steps shouldn't touch them, so this guarantees consistency.

bendavid:
Suggest to confirm that this is really giving the identical energy to the
one from the default 4 vector in the object...

sethzenz:
Yes.  I was assured this was true by EGM, and I also explicitly checked that getCorrectedEnergyError(regression_type) gave the same result as energy().  (Before any flashgg smearings, obviously.)

sethzenz:
For 100% clarity, regression_type in flashgg::Photon is fixed like this:

`static const reco::Photon::P4type regression_type = reco::Photon::regression1;`

PR #212, opened 2015-04-07
Topic rndcones and much more

musella:
- Added pluggable isolation algorithms in PhotonProducer
  -  Additional isolations stored as new data members in flashgg::Photon.
   - Used to implement random cone isolation.
   - Added getters for vertex 0 isolations

- Bug fix in IdleWatchdog for prescale = 1

- Dumpers:
  - Added ignoreCuts option in CutBasedDiPhotonObjectSelector (for N-M selection)
  - Added dumpOnly option to category dumper to store only a subset of variables

- Speed-up job splitting

### DIFF
--- a/DataFormats/interface/Photon.h
+++ b/DataFormats/interface/Photon.h
@@ -5,6 +5,9 @@
 #include "DataFormats/PatCandidates/interface/Photon.h"
 #include "DataFormats/PatCandidates/interface/PackedGenParticle.h"
 
+#include <map>
+#include <string>
+
 namespace flashgg {
 
     class Photon : public pat::Photon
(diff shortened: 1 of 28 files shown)


sethzenz:
I get this message while testing MicroAOD-making (with both this PR and #211): 

Error:  'userCustomize' not already registered

Things seem to run ok anyway.  Is this a concern?

musella:
The issue with the userCustomize should have not been a concern. I think that it come from the fact that the actual implementation of hasattr(obj,name) in python is

def hasattr(obj,name):
      try:
obj

sethzenz:
Ok, all is behaving fine now.

PR #214, opened 2015-04-09
synchronize btag discriminant name, restore untagged 2

sethzenz:
Fix a few bugs that were keeping some tags from showing up:

Untagged 2 was missing from the tag sorted (again!?) -- fixed.  

The btag discriminant was not synchronized between MicroAOD and Tag steps -- this is now fixed, and set from a central place: `MicroAOD/python/flashggJets_cfi.py`.  The Taggers are now all set up to get import this value.  However, note that if you run on PHYS14 MicroAOD the value should be changed to the discriminant that was used in that release.  Finally, note that our current discriminant is probably not the best one, it's just the one we can get to run correctly on 7_4_0_pre9!   There should soon be more general instructions and it will be updated then.


sethzenz:
For the record, to run consistently on PHYS14 MicroAOD, you should set flashggBTag to "combinedInclusiveSecondaryVertexV2BJetTags"

PR #218, opened 2015-04-14
fix astyle settings so linux lineend always imposed; ~15 files changed

sethzenz:


sethzenz:
A small number of files had ^M (aka /r) at the end of lines, i.e. the MSDOS standard rather than the linux standard.  The indent script was then adding ^M to new lines in these files, because it defaults to consistent end-of-line formatting.  This PR does three things:

1. Change Validation/scripts/indent.ast so it always removes ^M.  
2. Change Validation/scripts/flashgg_indent.sh so it can be run from any directory.
3. I ran Validation/scripts/flashgg_indent.sh and removed ^M from about 15 files that had them, which then show up as "invisible" diffs.

If you find that your text editor habitually adds ^M characters, you should run Validation/scripts/flashgg_indent.sh to remove them before committing.  If I notice that a PR adds ^M in the future I will ask you to do this in order to fix it.

PR #221, opened 2015-04-15
update catalog for v3

sethzenz:
This all worked.  It's clear the duplicate-removal functionality will make more sense if in the future we create new campaign name each time we rerun on the same datasets, so I will plan on that.

### DIFF
--- a/MetaData/data/HggPhys14/datasets.json
+++ b/MetaData/data/HggPhys14/datasets.json
@@ -53,6 +53,94 @@
             }
         ]
     }, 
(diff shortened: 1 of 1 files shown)


musella:
Rather than replacing the datasets in the old catalog, we should create a new one.

It would be enough to run the fggManageSample as

fggManageSample.py -C HggPhys14-Phys14MicroAODV3 import

And then redo the check to compute the weights.

sethzenz:
In this PR I have both datasets in the same catalogue.  Is it better to reject this PR and proceed as you suggest with a separate catalog instead instead?

musella:
Yes, I think so.

PR #226, opened 2015-04-23
DiMu and MuMuGamma EDM objects and dumpers

JunquanTao:
Create the dimu and mumugamma edm objects, then the dumpers and the validations


sethzenz:
In the long term we don't want these objects running in the standard sequence, but I think it is OK for now for debugging purposes.

PR #236, opened 2015-05-21
String-based photon energy scale adjustment module + example

sethzenz:
Create a tool that uses string parsing to create arbitrary photon systematic shifts in arbitrary bins.  Common binning can be given for many copies of the tool, which are each given their own range of applicability, so that different ranges can be correlated or uncorrelated as needed.  A simple example of this with uncorrelated barrel/endcap is implemented in Systematics/python/flashggPhotonSmear_cfi.py


sethzenz:
Will submit updated version later

PR #239, opened 2015-06-03
vtxId and vtxProb changes were included

InnaKucher:
VtxId and VtxProb MVA were retrained. 
Single leg conversion option was added to config file. 
VtxProb parameterization was included to the DiPhotonMVAProducer.

The code has to be tested when proper input files are available.


sethzenz:
MicroAOD and Tag tests run fine.

PR #240, opened 2015-06-05
workspace combiners, systematics, tag truth

sethzenz:
A lot of work all in one (rebased) commit:
* String-based photon energy scale
* pull in a copy of Benoit's workspace combiner
* A working (but imperfect) version of systematics dumping
* Truth information for tags in order to separate right/wrong vertex
   * Base class is propagated through for every tag except one, contains true vertex
   * More detailed example for VBF (pulling in some code I already had lying around)


sethzenz:
Also:
* I wrote proper copy constructors for all the tags and based the clone methods on them.
* Assorted spacing "fixes"
* all diphoton input tags now tracked. (this needs to be done more generally)

sethzenz:
New version of this PR will be made soon.

PR #245, opened 2015-06-25
Topic binned photon systs v8

sethzenz:
A lot of work all in one (rebased) commit:

* String-based photon energy scale
* Put in a copy of Benoit's workspace combiner
* A working (but imperfect) version of systematics dumping
* Truth information for tags in order to separate right/wrong vertex
   * Base class is propagated through for every tag except one, contains true vertex
   * More detailed example for VBF (pulling in some code I already had lying around)
   * I wrote proper copy constructors for all the tags and based the clone methods on them.
* Assorted spacing "fixes"
* all diphoton input tags now tracked. (this needs to be done more generally)
* String-based smearing (1d and 2d)
   * Handling of 2d in systematic producer
   * Handling of random engine in base class
* Read in of smearing and scaling files from globe
* Photon has label to allow for merging of workspaces for all systematics (which is not done yet)

Still to do:
   * Fix calculation of stochastic term - values don't look right currently
   * Unify all workspaces in dumper
   * Clean up workflow and configs, eventually


sethzenz:
Replaced by #248

Issue #27, opened 2014-08-04
Keep the link between the photon collection and photons stored into the diphoton object


OlivierBondu:

musella:
This should be straightforward to do with ShallowClones.

https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookParticleCandidates#ShallowCloneCands

sethzenz:
I've been looking at the ShallowClone solution, but I think there's a problem: we attach flashgg-specific information at the same time we make the new objects:

https://github.com/cms-analysis/flashgg/blob/7bbbf7a03cd172cbdc2ea6ce80d51efdb7ed4826/MicroAOD/plugins/DiPhotonProducer.cc#L105-L108

Is there a way to salvage this approach, or should we just put a reference back to the originals (if available), to go alongside our copies?

OlivierBondu:
as stated on the comment above the lines you quote, this information was needed for the preselection. The preselection, if we keep the current implementation using EDFilter (see https://github.com/cms-analysis/flashgg/blob/master/MicroAOD/python/flashggPreselectedDiPhotons_cfi.py), needs these in the photon candidate itself.

sethzenz:
So my conclusion is that it's easiest to simply put a ref to the original photons into diphoton separately.  Make sense, @OlivierBondu and @musella?

sethzenz:
Finally addressed by #251

PR #251, opened 2015-06-30
SinglePhotonView as component of DiPhoton

sethzenz:
Work by @rateixei with some interface tweaks by me.  Wraps each photon in a SinglePhotonView inside the diphoton, which provides a link to the original photon (addressing #27 ) and builds a cached photon with corrected P4 based on the diphoton vertex.  When saved, the cached photon is not included to save space in MicroAOD; it's rebuilt from the link to the original when next required.  However, for systematics it can be "made persistent", i.e. the changed photon is stored separately as part of a copied DiPhoton.  

In most cases, DiPhotons can be used exactly as they were before.  Methods like leadingPhoton(), subLeadingPhoton(), leadingView(), subLeadingView() are const pointers.  

Accessors with get in front return non-const references.  In order to protect the user from making objects larger when not intended, attempting to get a non-const reference before MakePersistent() is called causes a run-time exception; if photons are changed, computeP4AndOrder() must be called on the DiPhoton so that it returns correct information for its own momentum.  The MakePersistent and computeP4AndOrder methods are currently needed in only one place: Systematics/interface/DiPhotonFromPhotonBase.h    

At MicroAOD level, the impact is the following savings of space (collection, compressed size, uncompressed size):

Old: flashggDiPhotonCandidates_flashggDiPhotons__FLASHggMicroAOD. 9162.08 1679.16
New: flashggDiPhotonCandidates_flashggDiPhotons__FLASHggMicroAOD. 837.327 436.969

### DIFF
--- a/DataFormats/interface/DiPhotonCandidate.h
+++ b/DataFormats/interface/DiPhotonCandidate.h
@@ -11,9 +11,7 @@
 
 
 namespace flashgg {
-//    class SinglePhotonView;
 
-//    class DiPhotonCandidate : public math::XYZTLorentzVector //reco::LeafCandidate
     class DiPhotonCandidate : public reco::LeafCandidate
     {
     public:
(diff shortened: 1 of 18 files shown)


sethzenz:
Looks even better with higher stats:
flashggDiPhotonCandidates_flashggDiPhotons__FLASHggMicroAOD. 618.825 216.967

PR #255, opened 2015-07-02
rename UntaggedCategory to UntaggedTag for consistency with other tag…

sethzenz:
… names


sethzenz:
It seems I folded this into #256 by accident.

PR #256, opened 2015-07-03
MiniAOD: output list, customization; rename UntaggedTag

sethzenz:


sethzenz:
I will include some slides on how to configure the MiniAOD output with the python customization code on Monday.

PR #258, opened 2015-07-03
flashggMuon_weightMethodForLeptons_diphoMVAAutomation_LetptonEffSmearer

molmedon:
Added new variable which will store the weights for leptons (dummy scale factor values in python config)

Automation of diphomva Yacine&Louie style

LeptonEffSmear

Added flashggMuon class


sethzenz:
The following is not a problem for running the code or this PR, but it should be cleaned up.  It presumably originates from additional class defs added somewhere between CMSSW 740pre9 and 746.

Warning in <TInterpreter::ReadRootmapFile>: class  edm::Ptr<pat::Muon> found in libDataFormatsPatCandidates.so  is already in libflashggDataFormats.so

sethzenz:
Other todo's:
Initialize weight (to 1?) if not set otherwise.
Propagate weights through tags and into workspaces

sethzenz:
Code compiles and runs and produces sensible-looking output. Good work, thanks!  Will merge now.

PR #264, opened 2015-07-14
Update determineMatchType to work with cmssw 74X

nancymarinelli:


sethzenz:
Accepting for now but we should revisit the behavior here - the output is not logically the same as before.  In particular we call anything that's not a prompt photon from the Higgs a "fake" - is that what we want?

nancymarinelli:
Hello Seth,
Maybe I am confused. But why do you say that only photons from a Higgs are prompt and all others are fake ?
There is not request anywhere about the photon origin
And  isPromptFinalState()" would also include photons coming from QED ISR/FSR
(though photons from pi0/eta0/etc decays

In fact, for my analysis on the preselection, I had to use locally in my analysis the explicit request that the matched photons was Higgs's son.

  Nancy


______________________________________

Nancy Marinelli

Research Associate Professor
University of Notre Dame, IN, US

CERN, Bdg 40/3-A01, 1211 Geneva,

SWITZERLAND

Phone +41-22-76-70809, fax +41-22-76-78940

sethzenz:
If you read the code logic in this diff:

https://github.com/cms-analysis/flashgg/pull/264/files

You will see the following behaviors:

Before:  photon starts with a defaultType of Unknown. Under some conditions, it is set to Prompt; under others, it's set to Fake.  However, it can also remain as Unknown.

After: due to if-else structure, photon is *always* set to either Prompt or Fake.

sethzenz:
Ah, you are correct that we're not requiring the photon to be from the Higgs.  In any case, I accepted the code because it clearly suffices for existing studies.  But in the long-term we should remain aware that there are other possibilities, and that the code logic has changed.

PR #268, opened 2015-07-15
Couple of fixes

musella:
- switch debug flag off ini MicroAODCustomize.py
- add ::All to globa tags map in MetaData

The second needs to go in to properly run a production.

### DIFF
--- a/MetaData/work/campaigns/globalTagsLookup.json
+++ b/MetaData/work/campaigns/globalTagsLookup.json
@@ -1,5 +1,5 @@
 {
-    "Run2015B-PromptReco"                          : ["GR_P_V56::All"],
-    "RunIISpring15DR74-Asympt50ns_MCRUN2_74_V9A"   : ["MCRUN2_74_V9A::All"],
-    "RunIISpring15DR74-Asympt25ns_MCRUN2_74_V9"	   : ["MCRUN2_74_V9::All"]
+    "Run2015B-PromptReco"                          : ["GR_P_V56"],
+    "RunIISpring15DR74-Asympt50ns_MCRUN2_74_V9A"   : ["MCRUN2_74_V9A"],
+    "RunIISpring15DR74-Asympt25ns_MCRUN2_74_V9"	   : ["MCRUN2_74_V9"]
 }
--- a/MetaData/work/campaigns/globalTagsLookup.json
+++ b/MetaData/work/campaigns/globalTagsLookup.json
@@ -1,5 +1,5 @@
 {
-    "Run2015B-PromptReco"                          : ["GR_P_V56"],
-    "RunIISpring15DR74-Asympt50ns_MCRUN2_74_V9A"   : ["MCRUN2_74_V9A"],
-    "RunIISpring15DR74-Asympt25ns_MCRUN2_74_V9"	   : ["MCRUN2_74_V9"]
+    "Run2015B-PromptReco"                          : ["GR_P_V56::All"],
+    "RunIISpring15DR74-Asympt50ns_MCRUN2_74_V9A"   : ["MCRUN2_74_V9A::All"],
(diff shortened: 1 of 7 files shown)


musella:
In fact @crovelli made me notice that we should use ::All, as we are supposed to run on DB v2 for conditions.
Which means that we should replace 
`process.load("Configuration.StandardSequences.FrontierConditions_GlobalTag_cff")`
with
`process.load("Configuration.StandardSequences.FrontierConditions_GlobalTag_condDBv2_cff")`
in all cfg.py

sethzenz:
I don't think we want to set debug to false by default; we've not validated our ability to run all the code (esp. in the tag step) without those collections.  In particular, I think it will turn off the flashggPhoton collection, which we still need to keep.

musella:
Keeping debug on by default keeps a lot of unneeded products, including
rechits.

The flashgg photon collection should be in the default content, not in the
debug one.

Concering the tag step, the customization class is different, since
MetaData.JobConfig has to be used in that case.



On 15 July 2015 at 21:26, Seth Zenz <notifications@github.com> wrote:

> I don't think we want to set debug to false by default; we've not
> validated our ability to run all the code (esp. in the tag step) without
> those collections. In particular, I think it will turn off the
> flashggPhoton collection, which we still need to keep.
>
> —
> <https://github.com/cms-analysis/flashgg/pull/268#issuecomment-121720670>.
>



-- 
Pasquale Musella - ETHZ

musella:
Ah, I see now that there is a lot of content which is not included in the
default MicroAOD commands.

In particular:
- the flashggDiphotons should be moved into the default content, now that
they are lightweight;
- the PU info is needed for PU reweighting. A comment says that it is
"huge". How large is it?

Cheers,
P.


On 15 July 2015 at 23:26, Pasquale Musella <pasquale.musella@cern.ch> wrote:

> Keeping debug on by default keeps a lot of unneeded products, including
> rechits.
>
> The flashgg photon collection should be in the default content, not in the
> debug one.
>
> Concering the tag step, the customization class is different, since
> MetaData.JobConfig has to be used in that case.
>
>
>
> On 15 July 2015 at 21:26, Seth Zenz <notifications@github.com> wrote:
>
>> I don't think we want to set debug to false by default; we've not
>> validated our ability to run all the code (esp. in the tag step) without
>> those collections. In particular, I think it will turn off the
>> flashggPhoton collection, which we still need to keep.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/cms-analysis/flashgg/pull/268#issuecomment-121720670>
>> .
>>
>
>
>
> --
> Pasquale Musella - ETHZ
> http://musella.web.cern.ch/musella
>



-- 
Pasquale Musella - ETHZ

sethzenz:
(Discussed this via Skype; see #269 )

sethzenz:
I ran microAODstd.py with no extra customization on lxplus and got an issue with the GlobalTag:

`% cmsRun MicroAOD/test/microAODstd.py 
# Conditions read from  CMS_CONDITIONS  via FrontierProd 
----- Begin Fatal Exception 16-Jul-2015 10:59:05 CEST-----------------------
An exception of category 'ConditionDatabase' occurred while
   [0] Constructing the EventProcessor
   [1] Constructing ESSource: class=PoolDBESSource label='GlobalTag'
Exception Message:
Global Tag "MCRUN2_74_V9::All" has not been found in the database. from GTProxy::load 
----- End Fatal Exception -------------------------------------------------`

sethzenz:
See this thread: https://hypernews.cern.ch/HyperNews/CMS/get/commissioning/5840.html

Reading between the lines, it seems that the switch to condDBv2 and removing the All is not well documented (and perhaps not complete).  I think the problem may be that the "auto" tag pointed to "MCRUN2_74_V9::All" with the "All" included.  I am checking this and will fix if so.

sethzenz:
My guess is confirmed.  Hard-coding MCRUN2_74_V9 works and I will switch to that for now.  (In a following pull request.)

PR #270, opened 2015-07-16
move xml files to data directories

sethzenz:
All files we need at runtime have to be in `flashgg/*/data` (or `flashgg/*/interface`), because CRAB3 only "packs up" these directories - along with the compiled libraries.  Therefore the xml files for MVAs must be in the data directories.  This PR moves them and fixes a problem @matteosan1 emailed me about last night.


molmedon:
Hi Seth, as long as you also keep a copy in test/MVATraining/weights you need a dummy set of weights in here to be able to run the retraining python scripts .
On Jul 16, 2015, at 8:34 AM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


All files we need at runtime have to be in flashgg//data (or flashgg//interface), because CRAB3 only "packs up" these directories - along with the compiled libraries. Therefore the xml files for MVAs must be in the data directories. This PR moves them and fixes a problem @matteosan1<https://github.com/matteosan1> emailed me about last night.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/cms-analysis/flashgg/pull/270

Commit Summary

  *   move xml files to data directories

File Changes

  *   R Taggers/data/Flashgg_DiPhoDiJet_BDT.weights.xml<https://github.com/cms-analysis/flashgg/pull/270/files#diff-0> (0)
  *   R Taggers/data/Flashgg_DiPhoton_BDTG.weights.xml<https://github.com/cms-analysis/flashgg/pull/270/files#diff-1> (0)
  *   R Taggers/data/Flashgg_VBF_BDT.weights.xml<https://github.com/cms-analysis/flashgg/pull/270/files#diff-2> (0)
  *   M Taggers/python/flashggDiPhotonMVA_cfi.py<https://github.com/cms-analysis/flashgg/pull/270/files#diff-3> (2)
  *   M Taggers/python/flashggVBFMVA_cff.py<https://github.com/cms-analysis/flashgg/pull/270/files#diff-4> (8)

Patch Links:

  *   https://github.com/cms-analysis/flashgg/pull/270.patch
  *   https://github.com/cms-analysis/flashgg/pull/270.diff

sethzenz:
@molmedon -  I would separate development (e.g. training MVA's) from running the basic analysis sequence.  The locations and relative paths of code for development can be sorted out in a separate PR; I would favor putting it in Validation.  In any case, surely the scripts could point to a dummy in a different directory?

molmedon:
Yes, I think it is simpler to change the hardcoded path to point somewhere else, but the output should probably point somewhere else. I can go ahead and modify this .
On Jul 16, 2015, at 10:37 AM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


@molmedon<https://github.com/molmedon> - I would separate development (e.g. training MVA's) from running the basic analysis sequence. The locations and relative paths of code for development can be sorted out in a separate PR; I would favor putting it in Validation. In any case, surely the scripts could point to a dummy in a different directory?

sethzenz:
After the current PR, the scripts will still generate a weights directory with new xml files.  So the output will no longer go to the same path as later steps read it from.  That's ok, right?  In fact, I think it's better because we want to validate any output in detail before we commit it.

molmedon:
ah yes you are right. It is totally fine then.
On Jul 16, 2015, at 10:57 AM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


After the current PR, the scripts will still generate a weights directory with new xml files. So the output will no longer go to the same path as later steps read it from. That's ok, right? In fact, I think it's better because we want to validate any output in detail before we commit it.

PR #272, opened 2015-07-16
additional python configs to run diphoMVA retraining

molmedon:
additional python scripts for running diphoMVA retraining, forgot to add these in a previous pull request .

### DIFF
--- a/Taggers/python/DiPhotonMVADumperNew_cff.py
+++ b/Taggers/python/DiPhotonMVADumperNew_cff.py
@@ -1,0 +1,7 @@
+import FWCore.ParameterSet.Config as cms
+
+from flashgg.Taggers.DiPhotonMVATrainingDumpConfNew_cff import DiPhotonMVATrainingDumpConfNew
+
+DiPhotonMVADumperNew = cms.EDAnalyzer('CutBasedDiPhotonMVAResultDumper',
+                                **DiPhotonMVATrainingDumpConfNew.parameters_()
+                                )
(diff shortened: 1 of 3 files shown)


sethzenz:
These configurations look like cut-and-paste duplicates of other configurations for dumping the same classes.  I think there must be a more concise way to do this.  The worry is that if any configuration has to change, the more duplicates there are the more changes we'll have to keep track of.

molmedon:
ok, I was just following the model that Yacine and Louie were using. In principle this is not vital for the workflow, but trying to run the dumpers that output the files for the retraining won’t work without them. Do I just keep these in a dedicated branch instead ?



On Jul 16, 2015, at 6:35 PM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


These configurations look like cut-and-paste duplicates of other configurations for dumping the same classes. I think there must be a more concise way to do this. The worry is that if any configuration has to change, the more duplicates there are the more changes we'll have to keep track of.

molmedon:
ah! I think we should be able to put all of them into one config file ?


On Jul 16, 2015, at 7:37 PM, Manuel Alejandro Olmedo Negrete <manuel.alejandro.olmedo.negrete@cern.ch<mailto:manuel.alejandro.olmedo.negrete@cern.ch>> wrote:

ok, I was just following the model that Yacine and Louie were using. In principle this is not vital for the workflow, but trying to run the dumpers that output the files for the retraining won’t work without them. Do I just keep these in a dedicated branch instead ?



On Jul 16, 2015, at 6:35 PM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


These configurations look like cut-and-paste duplicates of other configurations for dumping the same classes. I think there must be a more concise way to do this. The worry is that if any configuration has to change, the more duplicates there are the more changes we'll have to keep track of.

sethzenz:
Accepting, there is a lot of redundancy in the dumpers, but not only in these.  We can clean up later.

Issue #162, opened 2015-02-11
MicroAOD size tuning


musella:
Current content (as of phys14 v2 production) is ~50% of MiniAOD.

We need to achieve a factor 2.5 reduction in order to meet the goal 1/5 goal.

I made a small excercie to see what could be done to meet the target.
https://musella.web.cern.ch/musella/higgs/flashgg/miniAOD.xls

Summary is that we need to:
- avoid photon info duplication in di-photon
- we should preselect leptons
- drop cluster information and store subset in photons
- only store photon ID info for vertexes associated with at least one di-photon pair
- reduce size of jet collection

sethzenz:
This is a nice study and I think it is all realistically achievable!  A few general principles, to remind everyone, as we work on this:
* All slimming and preselection that targets event size should be configurable. The slimming should be performed at separate steps from producers.
* Tag producers should NEVER count on particular objects having been preselected, with the exception of specifically-labeled preselected collections for the analysis, e.g. preselected diphotons.
The logic here is that we want to be able to "put back" extra information for validation and debugging, and we want to be sure that the behavior of the analysis does not change when we do this.

sethzenz:
Note that #27 is part of this task.

sethzenz:
Here is an absolutely huge item that was added at some point in the past:
PileupSummaryInfos_addPileupInfo__HLT. 14760.9 5335.92

sethzenz:
Here are the biggest offenders after the developments in PR #274, ignoring collections that we'll ultimately be able to remove once preselections are finalized.  This is for 50 ns ttH(gg): 

    flashggJets_selectedFlashggJets__FLASHggMicroAOD. 11386.4 1767.89
    flashggElectrons_selectedFlashggElectrons__FLASHggMicroAOD. 4440.14 1657.91
    PileupSummaryInfos_addPileupInfo__HLT. 4602.5 1640.32
    flashggPhotons_flashggSlimmedPhotonAndDiPhoton__FLASHggMicroAOD. 5046.86 1460.78
    flashggMuons_selectedFlashggMuons__FLASHggMicroAOD. 2974.24 1345.57
    recoCaloClusters_reducedEgamma_reducedEBEEClusters_PAT. 2964.51 1267.16
    recoSuperClusters_reducedEgamma_reducedSuperClusters_PAT. 3021.95 1028.47
    recoGenJets_slimmedGenJets__PAT. 3463.25 815.546
[...]

PR #274, opened 2015-07-17
new producer to build minimal photon collection from preselected dipho

sethzenz:
This is a (slightly tricksy) producer that takes in a diphoton collection and builds:
   * A new photon collection using only the photons pointed to by the diphotons
   * A new diphoton collection pointing to those photons
So the diphoton collection is the same size, but the photons can be smaller, for example after a preselection.

New collection is produced and saved, but no change is made in this PR to which collections are kept or which are read by the tag sequence.  We need the real preselection before we can finalize a switchover. 

Assuming a preselection with roughly the same "efficiency" as the current dummy, this reduces the file size from 16.1 kb / evt to 14.1 kb / evt.  Target is 9.6 kb / evt for the ttH(gg) sample I'm getting these numbers from.

### DIFF
--- a/DataFormats/interface/Photon.h
+++ b/DataFormats/interface/Photon.h
@@ -8,6 +8,7 @@
 
 #include <map>
 #include <string>
+#include <set>
 
 namespace flashgg {
 
@@ -22,6 +23,7 @@ namespace flashgg {
         virtual Photon *clone() const;
 
         void ZeroVariables();
+        void removeVerticesExcept( const std::set<edm::Ptr<reco::Vertex> > & );
 
         enum mcMatch_t { kUnkown = 0, kPrompt, kFake  };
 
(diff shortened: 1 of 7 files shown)


sethzenz:
Rereading #162, I realize this tool would be a good place to remove per-vertex info from the photons that is not relevant to the diphotons.  I'll update the PR with that addition today.

sethzenz:
Removed unneeded per-vertex info from: 

    std::map<edm::Ptr<reco::Vertex>, float> pfChgIso04_;                                                                                                                  
    std::map<edm::Ptr<reco::Vertex>, float> pfChgIso03_;                                                                                                                  
    std::map<edm::Ptr<reco::Vertex>, float> pfChgIso02_;                                                                                                                  
    std::map<edm::Ptr<reco::Vertex>, float> phoIdMvaD_;                                                                                                                                                                                                                                                                    
    std::map<std::string, std::map<edm::Ptr<reco::Vertex>, float> > extraChargedIsolations_;        

This saves a very 0.2 kB / evt (down to 13.9).  There is probably still more economizing to do inside the Photons - we need to review further what they store.

PR #280, opened 2015-07-20
Bug fix with 74X QGTagger

yhaddad:


sethzenz:
MicroAOD ok, Tag ok. Bon, d'accord.

PR #289, opened 2015-07-29
Multi-CHS Jet Collection handling +  Major modifications on the jet tools  

yhaddad:
This PR includes an implementation of a Muti-CHS Jet producer. This lead to a major modification on the jet tools. 

The jets are produced by :
* Add a Jet collection to process by running the `addFlashggJets` method : 
```python
from flashgg.MicroAOD.flashggJets_cfi import addFlashggJets
for vtx in range(0,5):
     addFlashggJets (process = process,
                     vertexIndex =vtx, 
                     doQGTagging = True,
                     label = '' + str(vtx))
```
* `vertexIndex` follows the same scheme as the define [here](https://github.com/cms-analysis/flashgg/pull/286)

These commands will automatically add 5 jets collections to the current process corresponding to the chosen vertex attached to the di-photon candidate
* All the Tag Producers that use jets are updated to use the Muti-CHS Jet collection
* An example of the Jet production can be found in `Validation/python/multi_jets_producer_test.py`


sethzenz:
Rejecting for rebase

PR #294, opened 2015-08-03
require full cluster info for selectedFlashggPhotons

sethzenz:


sethzenz:
Oops, I'm bad at rebasing.

PR #296, opened 2015-08-03
require full cluster info for selectedFlashggPhotons

sethzenz:
selectedFlashggPhoton now requires the full cuts from MiniAOD that are on both of these lines:

https://github.com/cms-sw/cmssw/blob/68b4ccfc9d6f7b0ae0980969622686957e61fc9a/RecoEgamma/EgammaPhotonProducers/python/reducedEgamma_cfi.py#L5-L6

Therefore starting with this collection we won't have photons with incomplete information.  Note that, for conceptual reasons and ease of remembering things, we might want to repeat these cuts in the preselection module once it's finalized.

### DIFF
--- a/MicroAOD/python/flashggObjectSelectors_cff.py
+++ b/MicroAOD/python/flashggObjectSelectors_cff.py
@@ -12,5 +12,6 @@ selectedFlashggMuons = cms.EDFilter("FLASHggMuonSelector",
 
 selectedFlashggPhotons = cms.EDFilter("FLASHggPhotonSelector",
     src = cms.InputTag("flashggPhotons"),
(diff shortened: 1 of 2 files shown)


musella:
I would rather have this in the diphoton preselection. Otherwise, we would never be able to see the photons that fail this preselection. 

My suggestion is that, in general, we should not select photons but diphotons.

sethzenz:
I think this is right.  We'll reduce confusion by getting rid of the selectedFlashggPhotons entirely.

As discussed, this means that dumping flashggPhotons will always give NaN's - so the correct procedure for dumping photons will really be to dump them via the diphotons.

PR #301, opened 2015-08-07
Few improvements to meta-data

musella:


sethzenz:
Accepting without tests, because:
(a) python only, will not interfere,
(b) I don't know what to test. ;)

PR #302, opened 2015-08-07
Removed explicit constructor in flashgg::Photon

musella:
The explicit copy constructor was introduced in 49ba012 @sethzenz `interface tweaks and cleanup, photon cl..`.

Not sure of what the reason was, but in general explicit copy constructors are a bad idea, as they need to be updated every time the class content changes.

In the current version a few data members were already missed in the copy. Only the extra-isolation variables were affected (so not the Hgg analysis), but this should be considered an example of what could happen.


sethzenz:
Agreed, this was a mistake on my part.

PR #311, opened 2015-08-12
Basic structure for CutAndClassBased systamtics dumping for Tags. Som…

ldcorpe:
…e work still to be done, mayeb worth rethinking structure of dumpers a little to avoid piling hack-over-hack

To be discussed: 
* Should we discard class-based dumpers entirely, and make it such that the cut-based dumpers also have the same features? Specifically, subcat could now be done using a cut on the value of categoryNumber() ?

* Want to avoid having to name the default collection "Nominal" as is currently done. Instead, could do something like : 
  for name in process.flashggTagSequence.moduleNames():
        module = getattr(process,name)
        if hasattr(module,"SystLabel"):
            module.SystLabel = "Nominal"
for the nominal process only when dumping systematics, thus leaving the systLabel empty when not doing systematics...

* Alongside previous point, could then avoid using "NOTFOUND" as default string here ? https://github.com/ldcorpe/flashgg/blob/topic-cut-and-class-based-systematic-dumper/MicroAOD/interface/CutAndClassBasedClassifier.h#L35-L36
At the moment this is needed because not having a systLabel returns "".

* Other comments?


sethzenz:
Can you look at this, @musella?  We should discuss this implementation, both in the short term and in terms of whether we can clean up and generalize in the long term.  

This is relevant also for systematics steering, @lgray.

sethzenz:
I think we might be able to solve some of these questions by replacing the string and hash with an enum, which we keep synchronized on the C++ and python sides, and have a string dictionary so systLabel() and setSystLabel() still work without actually storing a string.  That way we could have separate and explicit enums for NotAssigned, Nominal, and things like NotFound if needed.

ldcorpe:
Replaced systLabelHash by DiPhotonTagBase bool method hasSyst(label) as agreed during skype conversation. Additional tasks to be done in future are optimisation of what is dumped by only dumping some variables when running systematics, and implementing a flag for binned (rooDataset, TH1) or unbinned (rooDataHist, Tree) dumping.

PR #315, opened 2015-08-14
remove map from WeightedObject because it has bugs upon readback

### LINKED ISSUE #314: Remove all maps from EDM objects
Turns out that EDM object readback can fail for maps even within objects.  So we had WeightedObjects map<string,float> lose an entry mysteriously and cause a crash.  I'll fix WeightedObject but we should go back and look where else similar structures are used.

sethzenz:
Fixes #314 for WeightedObject.

### DIFF
--- a/DataFormats/interface/WeightedObject.h
+++ b/DataFormats/interface/WeightedObject.h
@@ -14,7 +14,7 @@ namespace flashgg {
     public:
         float weight( string key ) const;
         float centralWeight() const { return weight( central_key ); }
-        void setWeight( string key, float val ) { _labels.push_back( key ); _weights.push_back( val ); }
+        void setWeight( string key, float val );
         void setCentralWeight( float val ) { setWeight( central_key, val ); }
         bool hasWeight( string key ) const;
         void includeWeights( const WeightedObject &other );
@@ -25,7 +25,6 @@ namespace flashgg {
         vector<string> _labels;
         vector<float> _weights;
         static constexpr const char *central_key = "Central";
-
     };
 }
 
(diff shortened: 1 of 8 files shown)


DataFormats/interface/WeightedObject.h
```diff
@@ -13,40 +12,18 @@ namespace flashgg {
     {
 
     public:
-        float weight( string key ) const { return ( hasWeight( key ) ? _weights.at( key ) : -1. ); }
+        float weight( string key ) const;
         float centralWeight() const { return weight( central_key ); }
-        void setWeight( string key, float val ) { _weights[key] = val; }
+        void setWeight( string key, float val ) { _labels.push_back( key ); _weights.push_back( val ); }
```
lgray: As far as I can tell this does not do what you want. 
Any time you tried to set the value of an existing weight you'd append to the back (this would also break the sorting).

DataFormats/src/WeightedObject.cc
```diff
@@ -0,0 +1,55 @@
+#include "flashgg/DataFormats/interface/WeightedObject.h"
+
+namespace flashgg {
+
+    void WeightedObject::setWeight( string key, float val )
+    {
+        auto found_label = std::lower_bound( _labels.begin(), _labels.end(), key );
+        if( found_label == _labels.end() || *found_label != key ) {
+            _labels.push_back( key );
+            _weights.push_back( val );
```
lgray: This is still not quite the right behavior, you need to insert the value at the location where it should be in both vectors, otherwise the sort is broken.
Needs to be:
```
_weights.insert(_weights.begin()+std::distance(_labels.begin(),found_label),val); 
_labels.insert(found_label,key);
```

sethzenz:
Replaced by #330

PR #319, opened 2015-08-17
preselection with rho correction

camilocarrillo:
In this pull request:
0. A stand-alone configuration has been added to run just the preselection on top of the microAOD MicroAOD/test/testPreselection.py (for test porpuses or future changes in the preselection, no need to run again the rest of the algorithm)
1. The template with not tuned variables/thresholds but with the rho correction algorithm has been ported to MicroAOD/python/flashggPreselectedDiPhotons_cfi.py. 
2. A tool to compare diphoton collections (up to 4) has been added Validation/plugins/DiphotonPreselectionValidation.cc
3. This tool can be used for the validation of the preselection by comparing the diphoton collection produced after the preselection (with and/or without rho correction), here an example Validation/python/DiphotonPreselection_Validator.py
4. The variables and the thresholds need to be tunned before plugging the preselected collection to the rest of the analysis.

### DIFF
--- a/MicroAOD/python/flashggPreselectedDiPhotons_cfi.py
+++ b/MicroAOD/python/flashggPreselectedDiPhotons_cfi.py
@@ -1,82 +1,72 @@
 import FWCore.ParameterSet.Config as cms
 
-phoEffArea=cms.PSet( var=cms.string("abs(superCluster.eta)"), bins=cms.vdouble(0.,0.9,1.5,2,2.2,3), vals=cms.vdouble(0.21,0.2,0.14,0.22,0.31) )
-neuEffArea=cms.PSet( var=cms.string("abs(superCluster.eta)"), bins=cms.vdouble(0.,0.9,1.5,2,2.2,3), vals=cms.vdouble(0.04,0.059,0.05,0.05,0.15) )
+phoEffArea=cms.PSet( var=cms.string("abs(superCluster.eta)"), bins=cms.vdouble(0.,0.9,1.5,2,2.2,3), vals=cms.vdouble(0.16544,0.16544,0.13212,0.13212,0.13212) )
+#phoEffArea=cms.PSet( var=cms.string("abs(superCluster.eta)"), bins=cms.vdouble(0.,0.9,1.5,2,2.2,3), vals=cms.vdouble(1.,1.,1.,1.,1.) )
+#neuEffArea=cms.PSet( var=cms.string("abs(superCluster.eta)"), bins=cms.vdouble(0.,0.9,1.5,2,2.2,3), vals=cms.vdouble(0.04,0.059,0.05,0.05,0.15) )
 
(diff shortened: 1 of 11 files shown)


MicroAOD/python/flashggPreselectedDiPhotons_cfi.py
```diff
@@ -1,47 +1,82 @@
...
-                                      || subLeadingPhoton.chargedHadronIso()<0.3*subLeadingPhoton.pt()))               
+phoEffArea=cms.PSet( var=cms.string("abs(superCluster.eta)"), bins=cms.vdouble(0.,0.9,1.5,2,2.2,3), vals=cms.vdouble(0.21,0.2,0.14,0.22,0.31) )
+neuEffArea=cms.PSet( var=cms.string("abs(superCluster.eta)"), bins=cms.vdouble(0.,0.9,1.5,2,2.2,3), vals=cms.vdouble(0.04,0.059,0.05,0.05,0.15) )
+
+PreselectionVariables = cms.vstring(
+    "egChargedHadronIso", 
+    "egPhotonIso", 
+    "egNeutralHadronIso",
+    "hadTowOverEm",
+    "(?r9>0.8||egChargedHadronIso<20||egChargedHadronIso/pt<0.3?full5x5_sigmaIetaIeta:sigmaIetaIeta)",
```
musella: Here, I would use directly full5x5_sigmaIetaIeta. The conditional expression was there in the example because of the Phys14 MINIAOD bug.

MicroAOD/python/flashggPreselectedDiPhotons_cfi.py
```diff
@@ -1,47 +1,82 @@
...
+    "passElectronVeto"
+    )
+
+rediscoveryPreselection = cms.VPSet(
```
musella: Are those below the cuts tuned by Nancy? 

It looks like they are not...  If not, please put in the values (even if preliminary) that were presented.

musella:
This file also needs to be updated:
`MicroAOD/python/flashggZeeDiPhotons_cfi.py`

To avoid syncronization issues, one can import the cuts from flashggPreselectedDiphotons_cfi and the use the `ignore` or `invert` vectors.

`ignore = cms.vint32(list_of_cut_numbers_to_ignore)`
`invert = cms.vint32(list_of_cut_numbers_to_invert)`
`invertNtimes = cms.int32(2)`

Inversion logic is at:
https://github.com/cms-analysis/flashgg/blob/master/MicroAOD/src/CutBasedDiPhotonObjectSelector.cc#L25

camilocarrillo:
The update for the preselection with the optimized values prepared by Michael and Nancy have been addressed in https://github.com/cms-analysis/flashgg/pull/319

sethzenz:
Agreed in meeting that this will be redone

PR #321, opened 2015-08-18
TH1D for totWeights

andreypz:
This is to fix total weights calculation in fggCheckFile.py.
Other files are changed by the indent shell script requested to run before commit.


musella:
Hi Andrey,

thanks for the fix.

Could you also convert to TH1D the events histogram for consitency?

Thanks,
Pasquale

andreypz:
Ok, changed it via web edit. Should wok ;)

PR #324, opened 2015-08-18
Speedup indentation check

ferriff:
Got rid of some of the `find'.
Some (successful) tests done. Would benefit from some more.

### DIFF
--- a/Validation/scripts/flashgg_indent.sh
+++ b/Validation/scripts/flashgg_indent.sh
@@ -8,9 +8,7 @@ fi
 
 cd $CMSSW_BASE/src/flashgg
 
-find . -iname '*.C' > filelist
-find . -iname '*.cc' >> filelist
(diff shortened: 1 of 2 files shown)


sethzenz:
I tested also, this is a very nice improvement, thanks!

Validation/scripts/flashgg_indent.sh
```diff
@@ -8,9 +8,7 @@ fi
 
 cd $CMSSW_BASE/src/flashgg
 
-find . -iname '*.C' > filelist
-find . -iname '*.cc' >> filelist
-find . -iname '*.h' >> filelist
+git status -z | tr '\0' '\n' | grep "^M" | grep -E "(\.cc$|\.h$|\.C$)" > filelist
```
musella: This will miss the newly added files. Better grep -E "^(M|A)".

Issue #327, opened 2015-08-19
Check that user has a valid grid proxy before submitting with useAAA=1


sethzenz:
In fggRunJobs.py, we can (and should) check if the user has set up a grid proxy before submitting jobs that rely on having a grid proxy.  Just one more way to catch user error.

PR #329, opened 2015-08-19
check the validity of the proxy certificate when useAAA=True

ferriff:
should close issue #327.

### DIFF
--- a/MetaData/python/JobConfig.py
+++ b/MetaData/python/JobConfig.py
@@ -216,6 +216,7 @@ class JobConfig(object):
     # parse command line and do post-processing
     def parse(self):
 
+        import sys
         if self.parsed:
             return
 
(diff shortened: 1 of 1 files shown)


MetaData/python/JobConfig.py
```diff
@@ -224,6 +225,13 @@ def parse(self):
             self.readProcessIdMap(self.options.processIdMap)
         
         if self.useAAA:
+            import subprocess
+            tleft = int(subprocess.check_output(["voms-proxy-info", "--actimeleft"]))
```
musella: This code is normally run by cmsRun when it loads the configuration and it may be better not to do it for each job.

I would probably execute the command only if the

ferriff: only if..?

BTW, the induced latency is only about half a second per job (excluding
the time to import subprocess).

sethzenz:
Checking dryRun is a good idea.  But we absolutely don't want to wait for cmsRun for these checks, in case we're submitting batch jobs.  The jobs will fail and be resubmitted repeatedly, right?  Or should the dryRun already catch things like this?

sethzenz:
Discussion has dried up.  Priority for this is too low to sort it out so I am closing for the time being.

PR #334, opened 2015-08-20
dumpers - default variables, copy in MicroAODToWorkspace

sethzenz:


sethzenz:
Of interest to @lgray. Another round of tidying of the workspace dumpers.   We are testing this PR (and the adder too) on a big new processing round to make sure there aren't any little inconsistencies; so this PR might not be accepted until early next week.

PR #336, opened 2015-09-01
fix missing std:: for some of the `get' methods

ferriff:


sethzenz:
I note that L113 and L213 have "using namespace std" at limited scope, so these should not be technically needed unless there is a get in another namespace.  Is there a namespace collision here such that these change the code?  Or is this for clarity/style?

PR #340, opened 2015-09-14
Pdf uncertainty development

molmedon:
Still needs to rebase, some bogus commit. But here is the full machinery to incorporate pdf weights from cmssw.


sethzenz:
Looks good so far.  As discussed, this PR won't go in because there will be a new PR with improved compression (32 --> 16 bits).  When you make that new PR, can you do a bit of rebasing and also omit the following file changes?  I believe they are all either incidental or unrelated to this PR:

Taggers/plugins/DiPhotonMVAProducer.cc
MicroAOD/test/microAODstd.py
Taggers/test/simple_Tag_test.py

Taggers/test/MVATraining/test_diphodijet_compare/config.json
Taggers/python/DiPhotonMVATrainingDumpConf_cff.py
Taggers/test/MVATraining/test_diphodijet_compare/task_config.json
Taggers/test/MVATraining/test_diphodijet_training/config.json

PR #343, opened 2015-09-17
PDFWeightObjectProducer

molmedon:


molmedon:
This pull request has everything it needs :)

Each dataset may have different pdfsets. For this reason I include a separate cc file (PDFValidation.cc/pdfValidation_cfi.py) which will run over the lines of LHEEventProduct. This way we can easily check by hand which pdfsets are stored in the file. Once we have the name we can make a configuration (flashggPDFWeightObject_cfi.py) with some parameters that are needed. I did not include it into the main workflow (just added commented lines) because if the name of the pdfset is not found the script will crash (maybe this should change).

sethzenz:
Passes standard tests.  Size is very nice (on GluGluHToGG_M-125_13TeV_powheg_pythia8):

Uncompressed/Compressed: 147.58 36.952

TODOs for myself for later:
* Integrate into standard sequence, once we sort ouf which pdf sets should be present and should be used
* Integrate weights into systematics scheme - have workspace dumper read PDFs directly?

I'll leave this PR up 24 hours for comments, but all looks good to me.

PR #348, opened 2015-09-21
all flashgg classes have version set in xml to 10

sethzenz:


sethzenz:
N.B. If any flashgg/DataFormats classes are changed, a message like this will appear:

    error: class 'flashgg::DiPhotonCandidate' has a different checksum for ClassVersion 10. Increment ClassVersion to 11 and assign it to checksum 2243573479

If it does, you can edit flashgg/DataFormats/src/classes_def.xml to add a new "version" line with the new checksum and increment the version in the class name line.  https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Class_versioning  Get in touch with me if there are any issues with this!

PR #350, opened 2015-09-22
Few fixes to python:

musella:
- puppi removal customization to loop over all paths
- photon and diphoton cut-based dumper names


sethzenz:
I don't think this will create a conflict with the customization changes in #349 (although it probably removes the need for them).  But I'll accept this one first just in case...

PR #351, opened 2015-09-23
Topic lepton selection fix and Cutbased Dumpers fix

ldcorpe:
This PR contains two pieces of work:

*Work done to fix the Cutbased Dumpers since a previous commit of mine (to introduce CutandClass Based dumpers for systematics dumping) caused issues for the other dumper types.*

- Introduce `classname` option for addCategory(), which was previously taken from the label.
- Introduce `$CLASSNAME` substitution variable which can be used in the name template
- Fix logic to make sure Cutbased dumpers still functional alongside CutandClassbased dumpers

*Work done to fix lack of electrons-driven TTHLeptonic tags*
- Small bug in lepton Selection fixed (to do with an `||` instead of `&&`, making us lose all endcap electrons)
- Removed isolation requirement *temporarily* as the cut is about an order of magnitude below the returned valued from the isolation method. Needs to be looked at.
- Now returning some TTHLeptonic electron-driven tags, which were used to test the output of the "ElectronWeight" systematics. These appear to be working now (along with "MuonWeight") and arriving safely in the final workspaces as intended.

*Originally, this PR was going to include work to include IntLumi into workspaces, but this was removed following a comment from Pasquale about how to access the required info. To be included in a  later PR.*

### DIFF
--- a/Taggers/interface/CollectionDumper.h
+++ b/Taggers/interface/CollectionDumper.h
@@ -182,24 +182,37 @@ namespace flashgg {
         auto categories = cfg.getParameter<std::vector<edm::ParameterSet> >( "categories" );
         for( auto &cat : categories ) {
             auto label   = cat.getParameter<std::string>( "label" );
-            std::cout << "DEBUG CategoryDumper label " << label << std::endl;
+            //std::cout << "DEBUG CategoryDumper label " << label << std::endl;
             auto systLabel = cat.getParameter<std::string>( "systLabel" );
-            std::cout << "DEBUG CategoryDumper systlabel " << systLabel << std::endl;
+            //std::cout << "DEBUG CategoryDumper systlabel " << systLabel << std::endl;
             auto classname = cat.getParameter<std::string>( "classname" );
-            std::cout << "DEBUG CategoryDumper classname " << classname << std::endl;
+            //std::cout << "DEBUG CategoryDumper classname " << classname << std::endl;
             auto subcats = cat.getParameter<int>( "subcats" );
-            std::cout << "DEBUG CategoryDumper subcats " << subcats << std::endl;
+            //std::cout << "DEBUG CategoryDumper subcats " << subcats << std::endl;
             auto cutbased = cat.getParameter<bool>( "cutbased" );
-            //auto label2 = replaceString( label, "__" + systLabel, "" );
             auto name = replaceString( nameTemplate_, "$LABEL", label );
             name = replaceString( name, "$SYST", systLabel );
             name = replaceString( name, "$CLASSNAME", classname );
             auto key = std::make_pair(label, systLabel );
-            // if classname has been specified, use that instead. In general classname is not used (for now?)
-            // it was added in order to cater for the systematics case where the label is classname_syst
-            // but for the key, we only want the classname, so it is specified on its own as an additional optional argument.
-            if (classname != "") key = std::make_pair(classname, systLabel ); 
-            if (cutbased) key = std::make_pair("",label); 
+            bool classbased = (classname != "");
+            // the key is in general of the format <classname,cutname>
+            // for cutbased dumpers, the label is the cutname
+            if (cutbased && !classbased){ 
+            key = std::make_pair("",label); 
+            }
+            // for classbased, the label is the classname
+            if (classbased && !cutbased) {
+            key = std::make_pair(label,"");
+            }
+            //for cut-and-classbased, I guess the syntax should be to use the key <classname, label>
+            //if no systLabel is stored 
+            //if (classbased && cutbased && (systLabel=="")) key = std::make_pair(classname, label); 
+            //however, if a systLabel is specified, then this is the systDumper case, where the 
+            //cut is labelled by the systLabel. This could probably be merged with the case above in future. //FIXME
+            // for now this is the only application of cut-and-class dumper, so should be amended in future if needed.
+            if (classbased && cutbased) {
+                key = std::make_pair(classname, systLabel); 
+            }
             hasSubcat_[key] = ( subcats > 0 );
             auto &dumpers = dumpers_[key];
             if( subcats == 0 ) {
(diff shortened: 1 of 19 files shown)


Taggers/interface/CollectionDumper.h
```diff
@@ -182,12 +182,37 @@ namespace flashgg {
         auto categories = cfg.getParameter<std::vector<edm::ParameterSet> >( "categories" );
         for( auto &cat : categories ) {
             auto label   = cat.getParameter<std::string>( "label" );
+            //std::cout << "DEBUG CategoryDumper label " << label << std::endl;
             auto systLabel = cat.getParameter<std::string>( "systLabel" );
+            //std::cout << "DEBUG CategoryDumper systlabel " << systLabel << std::endl;
+            auto classname = cat.getParameter<std::string>( "classname" );
```
musella: I would actually make this optional.
std::string classname = "";
if( cat.exists("classname") ) { classname =  cat.getParameter<std::string>( "classname" ); }

Taggers/python/diphotonDumpConfig_cff.py
```diff
@@ -11,7 +11,7 @@
     categories = cms.VPSet(),
 
     workspaceName = cms.untracked.string("cms_hgg_$SQRTS"),
-    nameTemplate = cms.untracked.string("$PROCESS_$SQRTS_$LABEL_$SUBCAT_$SYST"),
+    nameTemplate = cms.untracked.string("$PROCESS_$SQRTS_$CLASSNAME_$SUBCAT_$SYST"),
```
musella: I think that $CLASSNAME should be taken out here, as the dumper is not class-based.

Taggers/python/dumperConfigTools.py
```diff
@@ -7,14 +7,17 @@ def addCategories(pset,cats,variables,histograms,mvas=None):
...
+   
     if subcats >= 0:
         catDef = cms.PSet(label=cms.string(label),
+				                  cutbased=cms.bool(cutbased!=None),
                           subcats=cms.int32(subcats),
                           variables=cms.VPSet(),
                           histograms=cms.VPSet(),
-													systLabel=cms.string(systLabel)
+													systLabel=cms.string(systLabel),
+													classname=cms.string(classname)
```
musella: Please set the default classname to None and replace the line above with.
if classname: catDef.classname = cms.string(classname).

And similarly for systlabel.

Taggers/python/dumperConfigTools.py
```diff
@@ -28,8 +31,15 @@ def addCategory(pset,label,cutbased=None,subcats=0,variables=[],histograms=[],mv
     if cutbased:
         cb = cms.PSet( cut=cms.string(cutbased) )
         if( label != "" ):
+            cb.name = cms.untracked.string(label)
+           # cb.name = cms.untracked.string(systLabel)
+        if( classname != "" ): #if classname AND cutbased, then this is
```
musella: Why not just using the nameTemplate instead of changing the labels here?

Taggers/python/diphotonDumpConfig_cff.py
ldcorpe: As discussed with @musella this is to be reverted to `$PROCESS_$SQRTS_$LABEL_$SUBCAT` and then nameTemplate will be changed to `$PROCESS_$SQRTS_$CLASSNAME_$SUBCAT_$SYST`  directly in the systematic dumping cmssw config.

Taggers/interface/CollectionDumper.h
ldcorpe: Changed as requested.

Taggers/python/dumperConfigTools.py
ldcorpe: Amending as suggested.

Taggers/python/dumperConfigTools.py
ldcorpe: In fact, here, I removed the need for l.36 etc by using the default label (for the systematic dumpers, composed of `classname_systlabel`) instead of the `systLabel` alone.

So the syntax is the same regardless of dumper type, as wanted.

In order to make this work however, I needed to add additional information into the cut for each dumper, as detailed in the message for the follow up-commit below.

ldcorpe:
Comments mentioned by @musella above have been addressed.
Notably, I now use the regular `label` instead of `systLabel` as the name for each dumper category, thus simplifying matters.

In order to make this work however, I needed to add additional information into the cut for each dumper (otherwise the simple cut on `hasSyst('systLabel')` passes for all the different Tag types, meaning that the  dumper category gets over-written by another Tag's version. e.g. we end up looking for a key of type `<UntaggedTag_0, VHEtTag_some-syst-variation>` and obviously the dumping failed.) 

The solution to this was to amend the Cut for the systematics to also cut on the Tag type. This is why I added the virtual method ` bool isTag(string)` to each Tag type. This allowed me to  make sure that each dumper category was unique not only to the Systematic but also to the Tag, thus avoiding the over-writing. Furthermore, this gave the opportunity to make further simplifications to the systematic dumper config: now instead of looping over systematics and having an `addCategory()` for each Tag, I loop over both Tags and Systematics with a single `addCategory()` which is populated which the correct information in each iteration of the couple loop.

ldcorpe:
Tested using:

`cmsRun Taggers/test/diphotonsDumper_cfg.py  inputFiles=file:/afs/cern.ch/user/m/musella/public/forLuie/diphotonsMicroAOD.root`

To make sure regular cut-based dumpers still work. Everything seems OK!

musella:
Ok. I think I managed to simplify things considerably in #362 and #363. Please @ldcorpe double check that the implementation works.

For the simple cut-based classification this are essentially reverted to the status prior to the introduction of class-and-cut classification.

I tested the systematics dumping with 
`cmsRun Systematics/test/MicroAODtoWorkspace.py  maxEvents=10 processId=ggh_120 inputFiles=/store/group/phys_higgs/cmshgg/sethzenz/flashgg/RunIISpring15-25ns/Spring15BetaV5/GluGluHToGG_M-120_13TeV_powheg_pythia8/RunIISpring15-25ns-Spring15BetaV5-v0-RunIISpring15DR74-Asympt25ns_MCRUN2_74_V9-v1/150922_093229/0000/myMicroAODOutputFile_1.root`

The main difference is that the nominal selection datasets are now named
`<process>_<sqrts>_<tagname>_<subcat>`
instead of
`<process>_<sqrts>_<tagname>_<subcat>_`

Which also matches what one would obtain from a simple class-name classification in the absence of the systematics loop.

@sethzenz, I would suggest closing this PR and merging the other two as soon as @ldcorpe gives his ok.

ldcorpe:
Hi all,

@musella 's update indeed provides the correct output to into the workspaces (checked the rooWorkspace contents manually) and is a much more elegant solution than what I was trying to put together today. Also I tested this with `cmsRun Taggers/test/diphotonsDumper_cfg.py inputFiles=file:/afs/cern.ch/user/m/musella/public/forLuie/diphotonsMicroAOD.root`, to make sure that the regular dumpers were OK, which they are by construction. 

Another advantage is that this new structure is much more easy to follow - thanks Pasquale!

There are some trivial changes to the finalFits inputs to be made, which I will take care of during the next full exercise.

@sethzenz - please go ahead and close this PR in favour or Merging Pasquale's.

One last note - to avoid situations where dumpers are broken by people's changes (although I imagine no further changes will be needed), maybe we should add  `cmsRun Taggers/test/diphotonsDumper_cfg.py inputFiles=file:/afs/cern.ch/user/m/musella/public/forLuie/diphotonsMicroAOD.root` (or similar) to the list of tests to be passed before a PR is accepted?

PR #353, opened 2015-09-23
handeling extra cmsRun arguments in prepareCrabJobs.py

yhaddad:
I added an option in `prepareCrabJobs.py` to include extra arguments. 
Example : If you want to send a crab job to run `cmsRun microAODstd.py puppi=2` you have to include the option `-e puppi=2` when you run `./prepareCrabJobs.py ... `

### DIFF
--- a/MetaData/work/prepareCrabJobs.py
+++ b/MetaData/work/prepareCrabJobs.py
@@ -116,8 +116,8 @@ parser = OptionParser(option_list=[
         make_option("-e","--extraPyCfgParam",
                     dest="extraPyCfgParam",
                     action="store",type="string",
-                    default="puppi=0",
-                    help="Extra python config parameters. Default: %default")
+                    default=None,
+                    help="Extra python config parameters. The arguments must be : -e 'arg1 arg2 ...' ")
         
         ]
                       )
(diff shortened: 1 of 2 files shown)


MetaData/work/prepareCrabJobs.py
```diff
@@ -109,7 +109,16 @@ def dumpCfg(options):
...
                     default="campaigns/globalTagsLookup.json",
-                    help="List of global tags to be used for data and MC. Default: %default")
+                    help="List of global tags to be used for data and MC. Default: %default"),
+        
+        # include additional parameters for cmsRun, such as the parameters from microAODCustomize
+        # the default here is puppi=0 (as an example) see microAODCustomize_cfg.py for more detail
+        make_option("-e","--extraPyCfgParam",
+                    dest="extraPyCfgParam",
+                    action="store",type="string",
+                    default="puppi=0",
```
musella: this should not be a default I think

MetaData/work/prepareCrabJobs.py
```diff
@@ -247,7 +256,12 @@ def dumpCfg(options):
         position = ProcessedDataset.find("-v")
         processedLabel = ProcessedDataset[:position]
         # print processedLabel
-
+        
+        # apprend extra parameters
+        if options.extraPyCfgParam:
+            #print "options.extraPyCfgParam:",options.extraPyCfgParam
+            replacements["PYCFG_PARAMS"].append(str("%s" % options.extraPyCfgParam))
```
musella: this would work only if the specified option is just one. I would be better to use "append" as action for extraPyCfgParam.
 
https://docs.python.org/2/library/optparse.html#standard-option-actions 

the call above would then become `replacements["PYCFG_PARAMS"].extend(map( lambda x: '"%s"' % x, options.extraPyCfgParam  )`

MetaData/work/prepareCrabJobs.py
musella: in fact, default should be None, or the if options.extraPyCfgParam would always be true.

yhaddad:
Thanks @Pasquale. Following your suggestion,  `prepareCrabJobs.py` can handle now many extra arguments.  The arguments must be included by ` -e 'arg1 arg2 ...' `.

PR #355, opened 2015-09-24
Same as #354

musella:
But for the master branch.


sethzenz:
My only question about this is whether we should keep these files in the main release:

    MetaData/data/RunIISpring15-25ns/datasets.json
    MetaData/data/RunIISpring15-50ns/datasets.json

After all, they will be 100% superseded and replaced when we do another production.

musella:
I think that we discussed this issue at the meeting on Monday.

Keeping old catalogs is important to compare with past results (and there is no limitation on the catalogs that can be committed).

Now, the issue is that the catalogs evolved in the meanwhile.

I suggest that we instead merge the prodBranch-Spring1BetaV5 branch in the master and close this PR.

The catalogs can be renamed as `RunIISpring15-(25)50ns-SpringBetaV5` just after the merge.

sethzenz:
I merged this PR.  I'll pick up the further changes separately.

PR #358, opened 2015-09-24
Topic gendipho and rechit flags

musella:
Notes:
- changes to flashgg::Photon are backward compatible, since the extra info is stored in PAT userInts, so no change in data format.
- I am not familiar with data format versioning so I did not provide it for GenDiPhoton.

### DIFF
--- a/DataFormats/interface/DiPhotonCandidate.h
+++ b/DataFormats/interface/DiPhotonCandidate.h
@@ -102,8 +102,11 @@ namespace flashgg {
         void setJetCollectionIndex( unsigned int val ) { jetCollectionIndex_ = val; }
         unsigned int jetCollectionIndex() const { return jetCollectionIndex_; }
 
+        LorentzVector genP4() const; 
     private:
 
+    private:
+        
         edm::Ptr<reco::Vertex> vertex_;
         int vertex_index_;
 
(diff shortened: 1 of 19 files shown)


sethzenz:
For future reference,  the procedure to add versioning to a class below.  I'm doing it as we speak in order to make these notes, so I'll make a separate PR once this one is accepted.

1. In classes_def.xml, turn this:

    <class name="flashgg::GenDiPhoton"/>

Into this:

   <class name="flashgg::GenDiPhoton" ClassVersion="10">
        <version ClassVersion="10" checksum="0"/>
    </class>

Note that version number 10 is the one to start with by convention in CMS.  Certain lower version numbers have a special meaning in ROOT and would break things.  Note also that "0" is a dummy checksum that we'll fix in the next step.

2. Build, and you'll get an error like:

    error: class 'flashgg::GenDiPhoton' has a different checksum for ClassVersion 10. Increment ClassVersion to 11 and assign it to checksum 3743016204

We do not want to increment the class version because we just created it, but we can...

3. Fill in the given checksum in the xml file:

    <class name="flashgg::GenDiPhoton" ClassVersion="10">
        <version ClassVersion="10" checksum="3743016204"/>
    </class>

4. Build again. All should be fine.

For the record, if an existing class is *changed*, then you should follow the instructions in the build error given in step 2 above.  That is, change to ClassVersion="11" in the first (class name etc) line, and add a new version line  for version ClassVersion="11" with the new checksum -- while keeping the version ClassVersion="10" and its checksum.  That way the code can recognize in the future which old version it's reading.

sethzenz:
See #359 for result of instructions above.

PR #359, opened 2015-09-25
add versioning for GenPhoton object

sethzenz:


sethzenz:
(also there was a long line that I resplit)

PR #364, opened 2015-09-30
Topic intlumi added to workspaces

ldcorpe:
This is a small PR which adds the `intLumi` into the final workspaces as an extra `RooRealVar`. The value of `intlumi` is either specified directly as an optional dumper config parameter or over-written by the metaData when using `fggRunJobs.py` using the value given by `targetLumi`.

@musella can you check this implementation is what you expected after our discussion on Monday?
Any comments or suggestions welcome!

Tested with `fggRunJobs.py --load test_jobs.json -d testMe -x cmsRun Systematics/test/MicroAODtoWorkspace.py maxEvents=100`
and `intLumi` is being pulled through correctly.
Also tested `cmsRun Taggers/test/diphotonsDumper_cfg.py inputFiles=file:/afs/cern.ch/user/m/musella/public/forLuie/diphotonsMicroAOD.root` and works fine


ldcorpe:
Replaced by https://github.com/cms-analysis/flashgg/pull/365 with fewer changes

PR #365, opened 2015-09-30
added IntLumi variable into workspaces. 

ldcorpe:
This is a small PR which adds the `intLumi` into the final workspaces as an extra `RooRealVar`. The value of `intlumi` is either specified directly as an optional dumper config parameter or over-written by the metaData when using `fggRunJobs.py` using the value given by `targetLumi`.

@musella can you check this implementation is what you expected after our discussion on Monday?
Any comments or suggestions welcome!

replaces https://github.com/cms-analysis/flashgg/pull/364

ested with fggRunJobs.py --load test_jobs.json -d testMe -x cmsRun Systematics/test/MicroAODtoWorkspace.py maxEvents=100
and intLumi is being pulled through correctly.
Also tested cmsRun Taggers/test/diphotonsDumper_cfg.py inputFiles=file:/afs/cern.ch/user/m/musella/public/forLuie/diphotonsMicroAOD.root and works fine

### DIFF
--- a/MetaData/python/JobConfig.py
+++ b/MetaData/python/JobConfig.py
@@ -166,6 +166,11 @@ class JobConfig(object):
                             wei *= xsec.get("br",1.)
                             wei *= xsec.get("kf",1.)
                             obj.lumiWeight = wei
+                    if hasattr(obj,"intLumi"):
+                        if isdata:
+                            obj.intLumi=self.targetLumi #FIXME is this where this info should come from for data?
+                        else:
+                            obj.intLumi=self.targetLumi
             
             for name,obj in process.__dict__.iteritems():
                 if hasattr(obj,"processId"):
--- a/MetaData/python/JobConfig.py
+++ b/MetaData/python/JobConfig.py
@@ -168,7 +168,9 @@ class JobConfig(object):
                             obj.lumiWeight = wei
                     if hasattr(obj,"intLumi"):
                         if isdata:
-                            obj.intLumi=self.targetLumi #FIXME is this where this info should come from for data?
+                            obj.intLumi= 0 # should not be used in final fits.
+                            # setting to 0 will cause error if someone tries to use
+                            #it for normalization downsteram
                         else:
                             obj.intLumi=self.targetLumi
             
--- a/Taggers/interface/CollectionDumper.h
+++ b/Taggers/interface/CollectionDumper.h
@@ -186,8 +186,10 @@ namespace flashgg {
         if( dumpWorkspace_ ) {
             ws_ = fs.make<RooWorkspace>( workspaceName_.c_str(), workspaceName_.c_str() );
             dynamic_cast<RooRealVar *>( ws_->factory( "weight[1.]" ) )->setConstant( false );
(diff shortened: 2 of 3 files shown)


Taggers/python/tagsDumpConfig_cff.py
```diff
@@ -6,7 +6,8 @@
     generatorInfo = cms.InputTag("generator"),
     processId = cms.string(""),
     maxCandPerEvent = cms.int32(-1),
-    lumiWeight = cms.double(1.0),
+    lumiWeight = cms.double(1.0),# Over-written by metaData if using fggRunJobs
+    intLumi = cms.untracked.double(1000.), # in /pb. to be over-written by metaData if using fggRunJobs
```
musella: Pleas add this also to the other dumpers, or else make the call on line 189 conditional on having the parameter (and take line 122).

`if( cfg.exists("IntLumi") ) { 
      RooRealVar* intLumi  = ...
}`

Taggers/interface/CollectionDumper.h
```diff
@@ -184,6 +186,9 @@ namespace flashgg {
         if( dumpWorkspace_ ) {
             ws_ = fs.make<RooWorkspace>( workspaceName_.c_str(), workspaceName_.c_str() );
             dynamic_cast<RooRealVar *>( ws_->factory( "weight[1.]" ) )->setConstant( false );
+            RooRealVar* intLumi = new RooRealVar("IntLumi","IntLumi",intLumi_*0.001,0,30000000);
```
musella: Please make this variable constant, as we don't want it to be floated in the fits. (

The downstream scripts can always enforce the constantness, but better to start on the right foot.

(also, you are not required to specify a range).

Btw, I see that for some reason the weight is not set constant and I am not sure why.

MetaData/python/JobConfig.py
```diff
@@ -166,6 +166,11 @@ def customize(self,process):
                             wei *= xsec.get("br",1.)
                             wei *= xsec.get("kf",1.)
                             obj.lumiWeight = wei
+                    if hasattr(obj,"intLumi"):
+                        if isdata:
+                            obj.intLumi=self.targetLumi #FIXME is this where this info should come from for data?
```
musella: For data the `IntLumi` parameter is not used, so we can even set it to 0 here, so that the downstream scripts get into error if they try to use it for normalization.

musella:
Hi Luie,

thanks a lot for the PR. I added a couple of inline comments.

Cheers,
P.

Taggers/python/tagsDumpConfig_cff.py
musella: Sorry, I got confused about the IntLumi parameter being tracked or untracked. Please disregard this comment: the code is fine as it is.

ldcorpe:
I've made a couple of amends as a result of suggestions by @musella:
*made `RooRealVar intLumi` a constant rather than floating parameter as suggested
*set `intLumi` to 0 for data, as suggested
*`intlumi` expected in 1/pb not 1/fb after all, so changed accordingly.

Please let me know if you have any further comments.

PR #369, opened 2015-10-07
Lighter Systematic Dumping with rooDataHists

ldcorpe:
It became clear after the last campaign that dumping systematic using full RooDataSets was too heavy, as we were running out of memory when adding files.
* Therefore, this PR adds the option to use RooDataHists instead, using the binning provided in the dumper config. The default is RooDataSets.
* The MicroAODtoWorkspace.py config now by default dumps systematics into RooDataHists, while the nominal collection is still a RooDataSet
* Space saved is substantial: testing on 30k events from a ggH sample, the size of an all-DataSet file is ~2.3MB while the size of the new syst-DataHist format is only 342KB. This is because to first order the size of a DataSet scales linearly with nEvents, while DataHists are ~constant.
* Also modified hadd_workspaces to correctly add RooDataHists together if they're present.  And also RooRealVars (like intLumi) are covered.

@musella please let me know if you have any comments about implementation, as this touches the Dumper framework.


sethzenz:
Tested with the new "standard" test list to make sure that dumpers are working for both workspaces and trees:
 cmsRun flashgg/MicroAOD/test/microAODstd.py
 cmsRun Taggers/test/diphotonsDumper_cfg.py
 cmsRun Systematics/test/MicroAODtoWorkspace.py

PR #370, opened 2015-10-09
Topic gammajet torebase2

martinamalberti:


sethzenz:
Tested that the code as a whole still runs sensibly.  As anticipated, no interference.

PR #372, opened 2015-10-13
customization tweaks for trigger and PDF uncertainties for next prod

sethzenz:
@musella @matteosan1 @molmedon @mplaner Can you double check the relevant parts of this code?


sethzenz:
I got no complaints at least so I'll go ahead and try starting the production with these changes.

PR #375, opened 2015-10-15
prepare for Spring15BetaV6

sethzenz:
This PR accumulates changes for Spring15BetaV6 running on reMiniAOD as described here:

https://hypernews.cern.ch/HyperNews/CMS/get/physics-announcements/3561.html

So far the 25ns portion is running, with the following not yet available:
   * The 2015C 25ns run
   * Some backgrounds: GJet, QCD_Pt-30toInf_DoubleEMEnriched_MGG-40to80, DiPhotonJetsBox
   * EGamma dataset does not exist (despite being in the json) - should it?


sethzenz:
Please note that the standard MicroAOD sequence is now changed in order to run on reMiniAOD, because one sample name changed.  There is a note in MicroAOD/test/MicroAODstd.py, indicating a line that must be uncommented in order to run on the older MiniAOD.

mplaner:
Hi Seth,
I noticed this when running over the 25ns run D.  It seems that "EGamma" is
split into "SingleElectron" and "SinglePhoton".  This is the case in the
current HLT configuration, so I believe it's expected.

For the tag-probe studies, I think also including the "SingleElectron"
would be useful.

--Michael

sethzenz:
@mplaner Thanks, I'll add SingleElectron for the next round of jobs.

PR #379, opened 2015-10-19
change kept pileup collection name

sethzenz:
Get collection name correct for reMiniAOD.  Previously I switched the module that reads the collection, but didn't update the output list.


sethzenz:
For the record this is about 5-10x smaller than before:

PileupSummaryInfos_slimmedAddPileupInfo__PAT. 1298.43 240.143

PR #381, opened 2015-10-19
Pdf validation fix

molmedon:
Deleted explicit declaration of copy constructor for PDFWeightObject and updated the checksum and class version in classes_def.xml


sethzenz:
Can you please put this line back (while keeping the adjacent changes)?
    <version ClassVersion="10" checksum="92077144"/>
That way the xml file will have a list of "old versions".  The line above will still say we're now on ClassVersion 11, but CMSSW will now recognize if it's looking at version 10 and give a better error message.  More info here: https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Class_versioning

PR #383, opened 2015-10-20
Pdf validation fix

molmedon:


sethzenz:
Tests ok, object filled now. Thanks!

PR #387, opened 2015-10-22
Integrate EGM TnP tool into flashgg

matteosan1:
Added few selector plugins to run TnP code with flashgg objects.
setup.sh script updated as well to checkout the basic TnP package.

Plugins location can be rediscussed, for the moment most of the relevant code is in flashgg/Validation but can be easily moved everywhere else.


sethzenz:
@matteosan1 This has conflicts and repeats of merge commits of previous PR's.  Can you rebase from the current upstream/master?

sethzenz:
This also touches the json files for Spring15BetaV5 and adds/removes whitespace.  Best to check for both of those things after the rebase.

PR #393, opened 2015-10-27
Finalize spring15 beta v7

sethzenz:
Last changes for Spring15BetaV7.  Should be identical to Spring15BetaV7 tag.


sethzenz:
Nope, it's not.   Will try again.

PR #395, opened 2015-10-27
change pdf weight config for workspace so default example runs smoothly

sethzenz:


sethzenz:
N.B. we are still working on actually getting the right numbers; this just means that the placeholder example runs without raising an exception because the vector's the wrong length.

PR #396, opened 2015-10-30
Electron and pdf upgrade

molmedon:
updated electrons such that the effective area is now centralized with recoegamma. For the pdf update, the script can now handle both madgraph and powheg for signal samples using pdfset number as the search parameter instead of pdfname . This may serve as a temporary solution or permanent depending on any updates from the generator group. This update will not work if both powheg and madgraph are used to produce one sample. Maybe someone can comment if this ever happens ?


molmedon:
forgot to modify pdf weight config file and there is also some unnecessary indentation, will close and rebase.

PR #399, opened 2015-11-03
move pdf weight obj into gen sequence so data customization removes it

sethzenz:


sethzenz:
This was already in signal customization.  The only issue is it crept into microAODstd.py.  so a new PR will be smaller.

PR #401, opened 2015-11-06
Spring15 beta v7 meta

sethzenz:
* JSON arithmetic to submit Spring15BetaV7 for Prompt MiniAOD after the reMiniAOD runs out
* Catalogue for those jobs, all of which are now run
* Fix to stray addition of pdf weights to default MicroAODstd.py that created problems in resubmission


sethzenz:
These changes will also be tagged (on top of Spring15BetaV7) as Spring15BetaV7_MetaV1

PR #402, opened 2015-11-08
Implemented PU reweighting

musella:
PU target distribution specified through VarParsing as putarget=<csv_represantation_of_data_distrib>.

The MC distributions is taken directly from the mixing module configuration in the release. 
Mapping between dataset names and mixing configuration  currently hardcoded in JobConfig.

Pu target must have same binning as distribution in the release (ie 52 bins between 0 and 52 at the moment).

To convert the output of pileUpCalc to the one needed as putarget, the following 5 lines script is sufficient.
``` 
#!/usr/bin/env python
import sys
import ROOT
fin=ROOT.TFile.Open(sys.argv[1])
pu = fin.Get("pileup")
pileup = map( pu.GetBinContent, xrange(1,pu.GetNbinsX()+1) )
print ",".join(map(lambda x: "%1.3g"%x, pileup))
```

Best agreement on Zee I obtained for mb_xsec=75.5mb

EBEB
nvtx_eb

EBEB Not HighR9
nvtx_eblowr9

EBEE
nvtx_ee)

EBEE HighR9
nvtx_eehighr9

EBEE Not HighR9
nvtx_eelowr9


sethzenz:
@musella I get a complaint about a missing parameter from the WeightsCountProducer, which we currently get via this recipe:

 git cms-addpkg CommonTools/UtilAlgos
 git cms-addpkg DataFormats/Common 
 git cms-merge-topic sethzenz:topic-weights-count-74X

Is an update required?

musella:
The commit above should fix it. 

(I had also modified the microAOD sequence in order to store the true PU distrib, but I I forgot to specify the binning).

sethzenz:
Thanks, tests all work.

PR #405, opened 2015-11-19
Fixes related to pu reweighiting 

### LINKED ISSUE #404: Bug in PU reweighting
Normalization is not preserved.

Need to investigate further.

musella:
- crash in dumpers when global variables are deactivated
- shift and normalization in pu reweighting

Fixes #404 

After the fix:
- normalization preserved (within ~1% on 100k events).
- reweighted distribution matches target one.


sethzenz:
Confirmed that diphoton and workspace dumping tests still work.

PR #406, opened 2015-11-24
adding preselection module

vtavolar:
Complete preselction with abs(eta)<2.5 and EB/EE gap removal, now in front of the tag sequence

### DIFF
--- a/MicroAOD/python/flashggPreselectedDiPhotons_cfi.py
+++ b/MicroAOD/python/flashggPreselectedDiPhotons_cfi.py
@@ -65,6 +65,9 @@ flashggPreselectedDiPhotons = cms.EDFilter(
         " && (leadingPhoton.hadronicOverEm < 0.08 && subLeadingPhoton.hadronicOverEm < 0.08)"
         " && (leadingPhoton.pt >30.0 && subLeadingPhoton.pt > 20.0)"
         " && (abs(leadingPhoton.superCluster.eta) < 2.5 && abs(subLeadingPhoton.superCluster.eta) < 2.5)"
+        " && (abs(leadingPhoton.superCluster.eta) < 1.4442 || abs(leadingPhoton.superCluster.eta) > 1.566)"
+        " && (abs(subLeadingPhoton.superCluster.eta) < 1.4442 || abs(subLeadingPhoton.superCluster.eta) > 1.566)"
+        " && (leadingPhoton.passElectronVeto) && (subLeadingPhoton.passElectronVeto)"
         ),
     variables = rediscoveryHLTvariables,
     categories = rediscoveryHLTcutsV1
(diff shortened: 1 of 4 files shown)


mplaner:
In general, I think this is okay.  For HLT studies and electron studies it's nice to have the preselection applied without the electron veto, so I'd rather have it somewhere else.  But it's probably safer to have it here for the average flashgg user.  I'd welcome more input.

ferriff:
i)  can the electron veto be made "switchable"?
ii) I would make all the cut values configurable, by chaining the cut string with values of variables in the filter, which would also make i) straightforward.
In this way one would always include flashggPreselectedDiPhotons from the default cfi and only make minimal changes via variable setting, rather than copy-pasting the whole cut string to change it.

sethzenz:
What I propose is that we stop saving preselected collections in MicroAOD in the future.  (We will save a size-reduced collection when needed that would be the common parts of the preselection between various studies, but this is not currently in high-demand.)  So we should run the preselection at the front of the tag sequence, for HLT and electron studies a different preselection should be defined to run in front, as needed.  Does this sound like a good plan?

mplaner:
Seth's plan seems reasonable to me.  And I think in general it's safer to do as Seth suggests and have a few preselections to choose from.  If we allow all the variables to be configurable, there's a good chance users will inadvertently tweak the preselection.  Having a "normal," "z->ee," "low-mass" preselection would all be useful.  Then we can also add in the mass cut.

sethzenz:
Normally I favor making everything configurable, but the preselection seems particularly prone to mix-ups.  Since it's very important to get it right, and issues can be subtle, I like the suggestion of having a short list of preselections (as listed by Michael) that are not particularly easy to reconfigure.  Is there a use case for running preselections that are different from these?  Of course we have to study changes in preselection, but that would usually be done with ntuples dumped from the unpreselected collection - not be tweaking the preselection module and checking the results.

sethzenz:
Hi @vtavolar, this doesn't work right if I merge it.  It depends on a couple of features you must have changed in Taggers/python/flashggDiPhotonMVA_cfi.py but did not include in the PR:

1. flashggDiPhotonMVANew removed, flashggDiPhotonMVAold added
2. flashggDiPhotonMVA must have been changed to load the preselected diphotons.  (Right now I get an assert error because the MVA collection size != preselected diphoton size.)

You can fix it if you like, or I'm happy to patch it myself.  But in the latter case can you confirm that I have understood the issues correctly?

vtavolar:
Hi Seth,
you are right. I forgot to commit the file containing the features you mentioned. It should work now.
Sorry again,

Vittorio

sethzenz:
Thanks!  I still need to do one or two tweaks to make the full tag test work out of the box.  So I'll close this PR and make a new one shortly.

PR #410, opened 2015-11-27
fix typo in sigmaIetaIeta cut for low R9 EE

sethzenz:
Per email thread, we've had a typo in the sigmaIetaIeta cut for the preselection since it was updated in September. @mplaner can you confirm this is the correct fix?


mplaner:
I confirm it is now correct.

PR #413, opened 2015-12-02
Diphoton MVA weights and misc.

vtavolar:
Hi Seth, all,

with this PR I would like to make available a set of the diphoton MVA weights. They correspond to the ones used to produce the plots showed in the last Hgg meeting. Together, I added the necessary changes in the DiPhotonMVAProducer to read the xml.

In addition to that, I include here some small features that might be worth merging as well:
- option to split weight into event-weight and lumi-weight, then dumped separately
- possibility to specify a signed integer (processIndex) for every dataset listed in the json loaded by fggRunJobs
- few more changes in the preselection, with a different treatment of the passElectronVeto variable. This implementation makes it easier if one wants to invert the cut, which can now be done with a customization instruction like

process.flashggPreselectedDiPhotons.variables[-1] = "-(passElectronVeto - 1)"

I had also put the cuts on pT/m directly in the preselection, do we want to have them there in the master?

Let me know which of these features you would like to keep or if I should remove them.

Bests,
Vittorio


sethzenz:
Hi @vtavolar, thanks for this!   I have a question and two requests:

1. what does processIndex do?
2. I think the default behavior of the preselection should be kept the same.  So making it more configurable is good, but I think the pt cuts should be set to the values they had before (not mgg dependent).
3. Can you please rebase to make the very long commit history simpler, and remove the merges from the master branch?  See https://twiki.cern.ch/twiki/bin/view/CMS/FLASHggFramework#How_to_rebase_before_PR for instructions.

vtavolar:
Hi Seth,

1. processIndex is added to the list of global variables dumped. It is helpful, for instance, in the training/test trees created by TMVA to distinguish to which process/sample an event belongs to.
2. ok, then I will remove the additional line containing the mgg-dependent pt cuts and I will keep the electron-veto stuff
3. sure, I apologize, thank for the instructions, I'll do it asap.

Vittorio

sethzenz:
Closing this PR, because a new one will be needed for the rebased version.  Thanks again!

PR #415, opened 2015-12-03
Diphoton MVA weights and misc. -- Rebase

vtavolar:
Hi Seth, all,

this PR is meant to supersede #413 with a cleaner commit history. 
The mass-scaling pt cuts in the preselection have been removed, as discussed.
Let me now if it is ok like this,

Thanks,
Vittorio


sethzenz:
We get messages like this on 864/1000 RunIISpring15MiniAODv2/VHToGG_M125_13TeV_amcatnloFXFX_madspin_pythia8 events:

Two qu

sethzenz:
I get messages like this on 864/1000 RunIISpring15MiniAODv2/VHToGG_M125_13TeV_amcatnloFXFX_madspin_pythia8 events:

MVA value is -999, dumping input variables
phoIdMva_SCRawE_         74.0082
phoIdMva_R9_             0.684753
phoIdMva_S4_             -nan
phoIdMva_covIEtaIEta_    0.0279462
phoIdMva_EtaWidth_       0.0212798
phoIdMva_PhiWidth_       0.0796546
phoIdMva_covIEtaIPhi_    999.9
phoIdMva_pfPhoIso03_     2.90063
phoIdMva_pfChgIso03_     26.0005
phoIdMva_pfChgIso03worst_26.0005
phoIdMva_ScEta_          -1.95724
phoIdMva_rho_            14.2058
phoIdMva_ESEffSigmaRR_   0

Two questions:

1. Is this a bad thing?  Is something misconfigured?
2. Should we mute this warning?

Thanks,
Seth

vtavolar:
Hi,

this is just something I added for debugging purposes and in principle it should not be there. I just wanted to check why some photons get a photon ID MVA score of -999, and my conclusion was that this is the return value from the TMVA reader when the value of input variables is outside of the expected range. In this case, and for all other events I checked, the culprit is S4, being -nan. I think it is not useful to have this printout here, so, if you agree, I will remove it. I don't know if we want to flag these events in any way, given that TMVA is already doing that with -999, but I'd like to hear other opinions.

Bests,
Vittorio

sethzenz:
Are these all events that would fail the preselection anyway?

sethzenz:
But yes, go ahead and remove it for now.

vtavolar:
Printouts removed with the latest commit.

I quickly checked in my trees and I have very few events (few tens in GJet and DiPho, 215 in the whole DoubleEG dataset)  where one of the two photons has photon ID =-999 even after the preselection module is applied. As far as the diphoton mva is concerned, there is a cut on the photon ID downstream that anyway discards these events.

PR #418, opened 2015-12-08
Fix getTokens() and some gen-specific code to work on data

ferriff:
* non trivial changes in VBFTagProducer.cc , please check


sethzenz:
Seems to behave and VBF tags with gen jets are produced.  Will accept for now, we'll double-check the details later.

PR #420, opened 2015-12-08
Scale and smearings for prompt reco 2015

ferriff:
* works from initial tests
* *pending validation* from Martina (expected some time tomorrow if everything smooth)

#### Notes

* scale for prompt 2015, values OK, errors as in Run1 (since Run2 stat. only)
w.r.t. the numbers in slide 3, the scale is computed as 1. / scale(MC) - 1.

* smearings for prompt 2015, values OK, errors as statistical of Run2 (since Run1 different parametrization)

* numbers from Fasanella et al., ECAL DPG 03/12/2015, https://cern.ch/go/zFH8


sethzenz:
Replaced by #420 due to merge conflicts with other PRs.

PR #422, opened 2015-12-09
machinery for sigmaE/E shift

vtavolar:
in the same style of PR #412 , machinery to shift sigmaE/E:
- Photon method to apply shift
- method used in systematics producer
- new syst added to MicroAODToWorkspace.py 

The shift has to be interpreted as a relative shift, so specifying a shift for sigmaE/E of 0.05 in Systematics/python/flashggDiPhotonSystematics_cfi.py corresponds to an applied shift of 5%.


ferriff:
So the shift is just relative, not in percentage, right?
Can you also add the documentation you wrote above to Systematics/python/flashggDiPhotonSystematics_cfi.py ?
Thanks.

vtavolar:
Yes, this is actually only relative, meaning 0.05 shifts the sigE/E by 5%. I will correct it and add it in the cfi.
Thanks,
Vittorio

PR #426, opened 2015-12-09
Topic develop rv syst technicalities

sethzenz:
Actual physics content, put in by @malcles, provides the correct RV/WV reweighting SFs.

This PR develops the machinery to apply systematics to the CombinedMVA plots, which had a few technicalities relating to collection templating and base class cloning and some tweaks required to the Systematics sequence in MicroAODToWorkspace.py.  I thought this was necessary because DiPhotonCandidate does not know the true vertex z, which was needed for determining the RV/WV reweight. In the longer term, several people expressed a preference to store the true vertex position in the DiPhotonCandidate (MicroAOD change) and convert this systematic back into a DiPhoton systematic.  But I keep this one for the Dry Run because it works and I already did it.

### DIFF
--- a/DataFormats/interface/DiPhotonCandidate.h
+++ b/DataFormats/interface/DiPhotonCandidate.h
@@ -103,7 +103,8 @@ namespace flashgg {
         unsigned int jetCollectionIndex() const { return jetCollectionIndex_; }
 
         LorentzVector genP4() const; 
-    private:
+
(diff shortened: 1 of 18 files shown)


sethzenz:
Note, I added a 400-7000 bin to avoid crashes.  It has a SF of 1 and an uncertainty of 0.  (N.B. we are not using the uncertainties on this for the dry run anyway; see the spreadsheet that has been circulated for details.)

PR #432, opened 2015-12-10
Hlt requirement

ferriff:
On top of PR #431 (easier not to mess up with python diffs).

Original feature:
* require hlt rediscovery path before anything else if processId == "Data"


sethzenz:
FYI I am testing these PRs (and mine as well) with new workspaces overnight.  I did the merging privately, and didn't do detailed tests, but if the new WSs look ok, then I can accept all tomorrow.

PR #433, opened 2015-12-10
photon scales on data and other ad hoc customization

sethzenz:
* do not run shifts up and down on non-signal, they are not used
   * run central photon scales on Data only

Not well validated, but I create this PR because it relates two things @ferriff  has time to look at:
1. Is the scale shift going in the correct direction, and does it improve Zee?
2. Refactoring the MicroAODToWorkspace.py configuration, which is growing rapidly in complexity as we put everything in.

### DIFF
--- a/Systematics/test/MicroAODtoWorkspace.py
+++ b/Systematics/test/MicroAODtoWorkspace.py
@@ -84,6 +84,17 @@ if customize.processId.count("h_") or customize.processId.count("vbf_"): # conve
 else:
     print "Data or background MC, so store mgg and central only"
     variablesToUse = minimalNonSignalVariables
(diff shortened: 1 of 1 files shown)


ferriff:
1. Scales and shifts are applied correctly (Martina checked by overpowering the scales). On the actual improvements there are still some checks to be done/understood, but the dry-run can proceed independently of that.
2. OK, thanks, I will look into that

Minor side comment: I find it a bit weird to set NoCentralShift to False to have it actually applied, I would just call it CentralShift and set it to True/False depending on the needs :) But we can take care of this when refactoring, I am fine having it like this for the dry-run.

sethzenz:
Yes, I can agree with that.  Maybe call it DoCentralShift (and switch true/false of course)

PR #439, opened 2015-12-17
no central values applied to data except photon scale

sethzenz:
Removing central shifts for data (except photon scale) as suggested by @martinamalberti. Previous behavior was incorrect.

@ferriff and @vtavolar, please comment if this commit would interfere with active work on the customization/steering.  If not, I think it's best to put it in so our starting point for that work is correct.


ferriff:
It slightly would, but never mind, much better to have the correct behaviour asap in the configuration we are exercising now, and then fix the re-worked customization when ready.

PR #443, opened 2015-12-20
run randomizer on (rerun) preselected collection

sethzenz:
Start the tagging and systematics sequences from preselected diphotons rather than "final" ones, because preselected is rerun and "final" is not.  Per email thread, both preselected and "final" will be removed from the next version of MicroAOD to avoid confusion, with preselection being run always at the start of tag/workspace building.  

Also, a few lines to debug which modules get diphotons from where.

### DIFF
--- a/MicroAOD/python/flashggRandomizedPerPhotonDiPhotonProducer_cff.py
+++ b/MicroAOD/python/flashggRandomizedPerPhotonDiPhotonProducer_cff.py
@@ -9,8 +9,8 @@ import FWCore.ParameterSet.Config as cms
 
 
 flashggRandomizedPerPhotonDiPhotons = cms.EDProducer("FlashggRandomizedPerPhotonDiPhotonProducer",
-                                    src = cms.InputTag("flashggFinalEGamma", "finalDiPhotons"),
-                                    outputCollectionName = cms.string("flashggFinalRandomizedDiPhotons"),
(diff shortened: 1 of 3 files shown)


ferriff:
Hi Seth, thanks. For the next version of MicroAOD we should probably run the randomizer on the single photon collection rather than the diphoton one. I can have a look at the necessary modifications tonight.
Cheers, Federico

sethzenz:
@ferriff Yes, if we do the randomizer in MicroAOD it has to be done before the diphotons are saved.  (One big reason is technical: otherwise the DiPhotons will "internalize" their photons and we will waste space.)  Probably best to do one of two things:

1. Put the randomizer directly into the PhotonProducer; or 
2. Run separately, and make sure to save only the photons with the random numbers included.

martinamalberti:
Thanks Seth. Shouldn't the input tag for the systematic producer be changed as well
https://github.com/cms-analysis/flashgg/blob/master/Systematics/python/flashggDiPhotonSystematics_cfi.py#L94
?

sethzenz:
Yes it should.  Looks like I missed that one when we changed over to rerunning the preselected, and it's probably actually the original source of the problem.

sethzenz:
OK, fixed – I think I missed that one repeatedly, sorry!  In fact before we run the next MicroAOD I will remove that final collection from everything.  Its only function is to save space, but between now and Moriond we don't need to save space, and we do need to minimize confusion.

PR #444, opened 2015-12-21
added hessian uncertainty calculation to flashgg pdf

molmedon:
This implements Josh's tool to calculate the uncertainties for the 100 replicas along the 60 eigenvectors of the hessian matrix. A couple questions remain, do we not have to divide the lhe->weights()[x] by lhe-> originalXWGTUP()  ?  What do we do with the alpha(s) variations ? At the moment they are not stored, only the 100 replicas are. Do we store them in a different vector ?

Besides these points, the machinery works well, it identifies if we are either using a madgraph or powheg generator and from here it sets hardcoded ranges for the pdf_id.

### DIFF
--- a/DataFormats/interface/PDFWeightObject.h
+++ b/DataFormats/interface/PDFWeightObject.h
@@ -17,8 +17,10 @@ namespace flashgg {
         ~PDFWeightObject();
         
         vector<uint16_t> pdf_weight_container;
-        
-        vector<float> uncompress() const;
+       
+	vector<uint16_t> alpha_s_container;
+ 
+        vector<float> uncompress( vector<uint16_t> ) const;
         
       };
  }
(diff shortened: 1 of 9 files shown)


sethzenz:
There are some technical issues with this PR.  @molmedon, do you have time today to look and update?  Below are a couple I fixed so far, but first here's one that I haven't fixed yet:

1. In `Taggers/interface/CollectionDumper.h`, and possibly other files, the uncompress() method is called without an argument.  But you changed the method so it requires an argument (and there is no default).  Did you forget to add a file or files that were changed?

And here are some I did manage to fix (but you could include if you update the PR to ensure it compiles).

2. There's a missing trailing slash in the added line for BuildFile.xml.  Instead of:

 <use name="PhysicsTools/HepMCCandAlgos">

You should have:

 <use name="PhysicsTools/HepMCCandAlgos"/>

3. The merge of Josh's code needs to be added to `setup.sh`, e.g. with:

 echo "Setting up PDF weight tool..."
 git-cms-merge-topic bendavid:pdfweights_74x

PR #450, opened 2015-12-23
Add PhotonRandomizedPhoton producer in the default MicroAOD sequence

ferriff:
* right after flashggPhotons
   * adjust defaults InputTags
   * add RandomNumberGeneratorService

MicroAOD production tested, output collection seems OK.

MicroAODtoWorkspace.py does not complain about missing products once sequences tweaked, although it fails with `binContents failed and would return a pair of empty vectors - 0th val:2.56093`. Since the error seems unrelated to this PR, I am submitting it anyway and I will debug in parallel the error.

### DIFF
--- a/MicroAOD/python/flashggDiPhotons_cfi.py
+++ b/MicroAOD/python/flashggDiPhotons_cfi.py
@@ -2,7 +2,7 @@ import FWCore.ParameterSet.Config as cms
 from flashgg.MicroAOD.flashggJets_cfi import maxJetCollections
 
 flashggDiPhotons = cms.EDProducer('FlashggDiPhotonProducer',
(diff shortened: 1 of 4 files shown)


sethzenz:
Bin contents issue is my fault.  It's fixed in 99.9% of cases in #447.  The issue is that, despite repeated efforts, we were still accidentally using the old preselection from MicroAOD.  When you truly run smearing before preselection (as we should), sometimes photons have eta beyond the binning...  Because we really do want to smear before preselection, the solution is to arbitrarily extend the last eta bins.  Extending to 3.0 fixes most events but there are still occasional issues; I'm confirming that just extending to 5.0 always solves the problem.

PR #457, opened 2015-12-29
Final changes for 1_0_0

sethzenz:
Final changes for flashgg 1_0_0:

* Remove preselection and finalEGamma from MicroAOD to avoid confusion
* Keep RecHits in data
* Keep exactly one photon and jet collection
* Fix bugs introduced in December PR's I made:
  * Generator PDF weight lookup logic change
  * Generator information added to DiPhoton causes gen module to run due to unscheduled mode even though collection is not opened in data in code: fix by completely removing generator module from process in data customization
* stupid_fix.py: script to keep job names <100 characters (due to crab feature); usage is in MetaData/README.md instructions
* Dumping tests, including PDF debug messages, random number lookup for jets, and eeBadSCFilter run on data
* jsons and campaigns for 1_0_0

Happy New Year!


sethzenz:
After this PR, the code is tagged as 1_0_0 and campaign is running:

http://dashb-cms-job.cern.ch/dashboard/templates/task-analysis/#user=Seth+Zenz&refresh=0&table=Mains&p=1&records=30&activemenu=2&pattern=&task=&from=&till=&timerange=lastWeek

The MC in this campaign could in principle be final for Moriond.  The data will need to be rerun when ReReco is available.

PR #462, opened 2016-01-05
Put DeltaPhi in absolute value

yhaddad:

PR #464, opened 2016-01-06
Absolute value of DeltaPhi

yhaddad:
Same as #462


sethzenz:
This is more concise and would fix the issue with the order of the absolute value and minimum.  But don't you still need an absolute value in defining dijet_dipho_dphi_?

PR #466, opened 2016-01-07
Jet updates for reading 1_1_0

sethzenz:
* Option to turn of central JEC recalculation; already ok for 1_1_0, so this is default.
* Some ground work for JES when no match is present (not done)
* Turn off debug messages
* Dumper for VBF Tag in order to check jet resolutions


sethzenz:
Part of 1_1_X

PR #467, opened 2016-01-08
Catalogue update: 1_1_X, all except MuonEG

sethzenz:
Everything is now there, DoubleEG luminosity is as expected, except for one very annoying MuonEG job.


sethzenz:
Final workspaces for 1_1_0.  Workspace submission scripts.  

Also, an extension of photon eta range for the diphoton systematics, because in one DY MC sample there's one photon with eta = 5.08 that caused a crash.  I don't think this is worth following up on because:

* It's super-rare.
* The photon will be removed by the preselection

But I note it in case anyone else would like to check.

sethzenz:
Replaced by #468

PR #468, opened 2016-01-10
Final 1_1_0 catalogues, +workspace script updates

sethzenz:
Final workspaces for 1_1_0, including all MuonEG and a confirmation that DoubleEG has exact expected lumi. Workspace submission scripts.

Also, an extension of photon eta range for the diphoton systematics, because in one DY MC sample there's one photon with eta = 5.08 that caused a crash. I don't think this is worth following up on because:

* It's super-rare.
* The photon will be removed by the preselection

But I note it in case anyone else would like to investigate.  The event number and file info are here:

```
----- Begin Fatal Exception 09-Jan-2016 23:31:20 GMT-----------------------
An exception of category 'Binning' occurred while
   [0] Processing run: 1 lumi: 178531 event: 44803593
   [1] Running path 'p'
   [2] Calling event method for module FlashggDiPhotonSystematicProducer/'flashggDiPhotonSystematics'
Exception Message:
binContents failed for method FlashggDiPhotonFromPhoton, label MCScaleLowR9EE, shiftLabel MCScaleLowR9EECentral, would return a pair of empty vectors - 0th val: 5.08347
----- End Fatal Exception -------------------------------------------------
09-Jan-2016 23:31:20 GMT  Closed file root://eoscms.cern.ch//eos/cms/store/group/phys_higgs/cmshgg/sethzenz/flashgg/RunIISpring15-ReMiniAOD-1_1_0-25ns/1_1_0/DYJetsToLL_M-50_TuneCU\
ETP8M1_13TeV-amcatnloFXFX-pythia8/RunIISpring15-ReMiniAOD-1_1_0-25ns-1_1_0-v0-RunIISpring15MiniAODv2-74X_mcRun2_asymptotic_v2-v1/160105_222534/0000/myMicroAODOutputFile_82.root
```


sethzenz:
Tagged as 1_1_1

PR #472, opened 2016-01-13
restored some old functionality to be able to handle mc samples outside of flashgg

molmedon:
In this pull request some old functionality was restored to handle recent cases for samples outside flashgg. Four flags were added to control different parts of the code which are format specific. 

For flashgg samples the flags for the scale, alpha_s and "flashgg_flag" need to be turned on in the python file and the "pdfset_flag" off .

For different mc samples (depending on the lhe txt format) the "flashgg_flag" needs to be turned off, alpha_s (depending on the pdf set) needs to be turned off, scale turned on and "pdfset_flag" on .

### DIFF
--- a/MicroAOD/plugins/PDFWeightObjectProducer.cc
+++ b/MicroAOD/plugins/PDFWeightObjectProducer.cc
@@ -35,6 +35,7 @@ namespace flashgg {
         string tag_;
         string delimiter_1_;
         string delimiter_2_;
+        string delimiter_3_;
         void beginRun( edm::Run const &, edm::EventSetup const &iSetup );
         vector<int> weight_indices;
         vector<int> alpha_indices;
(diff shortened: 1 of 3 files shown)


sethzenz:
Hi @molmedon, just to be clear:  you expect all four SM signal samples (and all SM backgrounds) to work correctly with the default configuration without anything special?

molmedon:
yes, this will work as it was before for SM signal samples for higgs to two photons.
On Jan 14, 2016, at 9:29 AM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi @molmedon<https://github.com/molmedon>, just to be clear: you expect all four SM signal samples (and all SM backgrounds) to work correctly with the default configuration without anything special?

ferriff:
Bare with me, but how did you test it now that all the relevant part of the analyzer PDFValidation.cc seems commented out?

molmedon:
I did that so that users can easily check the format of the LHERunInfoProduct text output before they start running jobs. I did not check if the member vectors were actually being filled directly. The same functionality that was working before for our samples remains, it is just flagged now since it won’t work for Rafael and Amina for example. For them I had to restore/add some code that was used before in order for them to be able to run jobs with pdf uncertainties. I did do some lower level checks, but I can quickly check that it is indeed  filling the member vectors :)

I do recommend that we keep PDFValidation.cc<http://PDFValidation.cc> as is so that users can check and not have their jobs crash.

On Jan 14, 2016, at 2:19 PM, ferriff <notifications@github.com<mailto:notifications@github.com>> wrote:


Bare with me, but how did you test it now that all the relevant part of the analyzer PDFValidation.cc<http://PDFValidation.cc> seems commented out?

ferriff:
Yes, a check that everything works exactly as before is advisable, thanks for this.

Concerning the crash: as I said in another thread, I think you should protect the code against it and issue warnings instead.

sethzenz:
Moved to 1_3_1 because not needed for initial SM signal production (which we want to start tonight). We'll get this in early next week so other analyses can do signal samples.

molmedon:
Hi I updated this pull request to prevent crashing and print a warning when txt format is invalid.


On Jan 15, 2016, at 4:24 PM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


Moved to 1_3_1 because not needed for initial SM signal production (which we want to start tonight). We'll get this in early next week so other analyses can do signal samples.

sethzenz:
Replaced by #501

PR #481, opened 2016-01-14
fix tth taggers

malcles:
Hi Seth,
here are some improvements/fixes for ttH taggers and lepton selection.
Finally I did not attach std egamma ids to the electron, as it seemed maybe not a so good idea,
but we have the std egamma id cuts in the taggers config for now.
Cheers,
Julie


sethzenz:
This PR turns off the VBFTag!  I'll leave it now as a I rebase to avoid confusion, but we must go back and fix it.

malcles:
Yes sorry Seth, this was the purpose of the question in Saranya’s talk on monday but I just realized we did not talk about it finally.
We were wondering in which order should the tags go. From the legacy paper, TTHleptonic should go first and
 then VBF dijet tag 0-2 and then TTHHadronic. I am sorry, can you put back VBF in between the two?
I have tested the code only with this configuration and I have not addressed how many ttH events
  end up in the VBF that way.

Julie



On Jan 19, 2016, at 17:53, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


This PR turns off the VBFTag! I'll leave it now as a I rebase to avoid confusion, but we must go back and fix it.

sethzenz:
See #499

PR #496, opened 2016-01-18
Pdf final development rebase 1 2 x

sethzenz:
Replaces #472.  N.B. this is a pull to the 1_2_X branch, which will now have to be kept separate from the master.  This is because 1_2_X will process CMSSW 74X MiniAOD and the master will process 76X MiniAOD, and the experts recommend we keep the versions consistent.


sethzenz:
We are going to avoid maintaining two separate branches, so I abandon this PR.

PR #497, opened 2016-01-18
Fix CollectionDumper in 76X

simonepigazzini:
details:

- Compatibility between FullFramework and FWLite kept only where critical (i.e. so that flashgg compiles)
- getByLabel substituted with getByToken in the CollectionDumper workflow.

### DIFF
--- a/MicroAOD/interface/GlobalVariablesComputer.h
+++ b/MicroAOD/interface/GlobalVariablesComputer.h
@@ -2,8 +2,16 @@
 #define flashgg_GlobalVariablesComputer_h
 
 #include "FWCore/Common/interface/EventBase.h"
+#include "FWCore/Framework/interface/Event.h"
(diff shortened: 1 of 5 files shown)


sethzenz:
Under review by @ferriff to see if FWLite can be preserved

simonepigazzini:
I think that two wrapper are needed so that FWLite can still calls a method with EventBase and getByLabel, while FullFW calls one  with Event and getByToken.

I did something like that in one of the helper utilities that are modofied in the PR, but maybe there is a cleaner way...

Cheers,

Simone 


Inviato da BlueMail



Il giorno 19 gen 2016 18:55, alle ore 18:55, Seth Zenz <notifications@github.com> ha scritto:

sethzenz:
With #505 checked in this PR should no longer be necessary.  Please test the upstream/master to confirm all works as expected now.

PR #494, opened 2016-01-18
incomplete catalogues and campaigns omnibus

sethzenz:
Should supersede #493.  Please check, @ferriff

PR #493, opened 2016-01-18
Some more 74X samples (SingleEle + ttH MC backgrounds)

ferriff:
* json for the production of SingleEle2015D (74X re-reco) and some ttH backgrounds (TTG, TTJ, TGJ, WG, ZG).
   * README updated accordingly

PR #500, opened 2016-01-19
copy only catalogue changes from topic-incomplete-catalogues-74xrerec…

sethzenz:
…o-76xrerero-74xtthbkg


sethzenz:
Replaces #494 and includes #493

PR #502, opened 2016-01-20
Combined data ReReco catalogue

sethzenz:


sethzenz:
NB The new combined catalogue is SingleElectron and DoubleEG for 2015D only.  If 2015C is needed please let me know.  (I might run it anyway, but no promises unless someone asks.)

sethzenz:
Ah, 2015C wasn't run for the 74X ReReco.  So we are stuck with what we have here!

PR #505, opened 2016-01-20
Fix consumes dumpers

ferriff:
Should maintain FWLite compatibility everywhere.
Adjust dumpers and the GlobalVariables stuff.

This PR should supersed #497 , @simonepigazzini please check.

It needs PR #504 for testing.
Tested with the standard `MicroAOD/test/microAODstd.py` and `Taggers/test/diphotonsDumper_cfg.py`. I don't have examples for FWLite, but  maybe @andreypz does and could make a quick test.

### DIFF
--- a/MicroAOD/interface/GlobalVariablesComputer.h
+++ b/MicroAOD/interface/GlobalVariablesComputer.h
@@ -2,8 +2,12 @@
 #define flashgg_GlobalVariablesComputer_h
 
 #include "FWCore/Common/interface/EventBase.h"
+#include "FWCore/Framework/interface/ConsumesCollector.h"
 #include "FWCore/Utilities/interface/InputTag.h"
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
+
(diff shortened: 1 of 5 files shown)


simonepigazzini:
@ferriff : I get seg-fault from CollectionDumper::_init() while running photon dumper, I'm trying to fix it...

ferriff:
Which photon dumper? The one indicated in the PR tested fine.
Federico

simonepigazzini:
Ciao Federico, as stated  in the other PR the issue is with global variables. Anyway this is the dumper:
https://github.com/musella/diphotons/blob/master/Analysis/config/photonIdDumper.py

simone

ferriff:
I was asking for the photon dumper to understand why the one used for
testing missed the issue...

simonepigazzini:
Looking in details the one that you cite also has DumpGlobalVariables set to True, so either you used a modified version or we are missing something. I'm sure that the fix provided in PR #508 make sense but I'm not sure if an unitialized bool is set to True or False, I think false...

simone

PR #508, opened 2016-01-21
Fix GlobalVariableComputer definition in CollectionDumper

simonepigazzini:
@ferriff I changed just a couple of lines. We need to instantiate dumpGlobalVariables_ before the if statement that checks for it in order to create the GlobalVariableComputer, otherwise we end up with a seg-fault.


ferriff:
Fine with me, thanks.

PR #512, opened 2016-01-22
fix-taggers-seq-and-tracked-params

malcles:
Hello Seth,
here is the pull request: 
- untrackedParameter moved to normal parameters in TTH taggers plus VHLoose and VHTight taggers, and the corresponding config flashggTags
- add back VBF tag in the Tag sequence and in the tag sorting, move it to after ttHHadronic (see email to flashgg for the explanation)
- some uniformisation of parameters names between taggers + change irrelevant names
- getByLabel removed from ElectronProducer (not working anymore in 76X apparently).

Cheers,

Julie


sethzenz:
Looks good, thank you!

PR #517, opened 2016-01-25
Changes for latest workspaces

sethzenz:
* Added to workspace dumper: TTHLeptonic, TTHHadronic
* Re-added to workspace dumper: electron + muon systematics
   * Change placeholder electron systematics to 1 (with large uncertainties for testing)
* Diphoton systematics eta range extended to 6 
   * Systematics run before preselection
   * eta(gg) > eta(1), eta(2) is possible if Higgs going down beampipe and photons are opposite phi
   * There is one known DY MC event with eta = 5.08 that crashes without this change
* Fix typo from #511 in which 0-weight bin goes to 30 for sublead photon
* Job scripts
   * Use 1_2_0 Filtered TTH samples
   * Use 74X ReReco DoubleEG
* Hadd script tweaks
   * No longer needs to merge all signal WS (files too big, ok due to updates to final fit code)
   * Merge all data


sethzenz:
For the record my reasoning about etas is flawed because we don't use the diphoton eta.  The rare crash in DY comes from a bizzare single photon with eta ~ 5.08.  In any case the preselection will get rid of it

PR #519, opened 2016-01-27
Correct CRLF -> LF for Taggers/python/flashggTags_cff.py

ferriff:


sethzenz:
Thanks!  For the record, as we discussed, at some point we should look to enforcing this in flashgg.  

For everyone's information, CRLF = "\r\n" and LF="\n".  Most files in flashgg use, and should continue to use, just the LF.  The problem is that different conventions in different text editors cause very messy diffs.

PR #520, opened 2016-01-27
 add std electron ids

malcles:
Hello Seth,

adding electron std egamma ids as said on monday, ie without modifying microAOD format for now,
 which is not very elegant but can be changed in the future (i can do this next week eventually if needed).
Default ttH tags use std egamma cut based loose id for electrons and deltaR cuts between leptons and photons of 0.4. No cut on diphoMVA yet in the default config. To be discussed at tomorrow working meeting.

Cheers,
Julie


sethzenz:
Thanks for this, @malcles.  There are some issues with this PR that ideally could be cleaned up (e.g. new tag with rebase and fixes):

* The many lines changed in Taggers/python/flashggTags_cff.py probably reflect a change from end-of-line from "\n" (LF) to "\r\n" (CRLF).  This was a previous issue that was fixed in #519, and I think this PR puts it back.  That should be fixed, and it should be possible automatically with the autocrlf option in ~/.gitconfig – @ferriff can comment if further details are needed.
* The EgammaIDs method re-encodes the standard EGM cuts.  This is dangerous and ideally would be replaced via the information in the MiniAOD (as is being discussed on https://hypernews.cern.ch/HyperNews/CMS/get/egamma/1676/1/1/1.html )   Ideally you could change the PR to use a more official setup, but if this would take too long than an assurance that all the numbers are checked very carefully will suffice for now (in my view).
* At github.com/cms-analysis/flashgg/pull/520/files#diff-cf1dab9e79d0d37cfe2e01d306d015c6R378 and vicinity, many lines of calculations are made to manipulate 4-vectors.  These could be shortened and made less error prone by using TLorentzVector methods like, for example, Eta() and Px()

Let's converge today on which cleanup is quickly achievable so we can proceed.

malcles:
Hi Seth,

first and last bullets will be done by tonight.
I will start from scratch and make a new PR, is it ok like that?

Second bullet, i can implement it next week in a cleaner way.

Julie



On Jan 28, 2016, at 12:17, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for this, @malcles<https://github.com/malcles>. There are some issues with this PR that ideally could be cleaned up (e.g. new tag with rebase and fixes):

  *   The many lines changed in Taggers/python/flashggTags_cff.py probably reflect a change from end-of-line from "\n" (LF) to "\r\n" (CRLF). This was a previous issue that was fixed in #519<https://github.com/cms-analysis/flashgg/pull/519>, and I think this PR puts it back. That should be fixed, and it should be possible automatically with the autocrlf option in ~/.gitconfig – @ferriff<https://github.com/ferriff> can comment if further details are needed.
  *   The EgammaIDs method re-encodes the standard EGM cuts. This is dangerous and ideally would be replaced via the information in the MiniAOD (as is being discussed on https://hypernews.cern.ch/HyperNews/CMS/get/egamma/1676/1/1/1.html ) Ideally you could change the PR to use a more official setup, but if this would take too long than an assurance that all the numbers are checked very carefully will suffice for now (in my view).
  *   At github.com/cms-analysis/flashgg/pull/520/files#diff-cf1dab9e79d0d37cfe2e01d306d015c6R378<http://github.com/cms-analysis/flashgg/pull/520/files#diff-cf1dab9e79d0d37cfe2e01d306d015c6R378> and vicinity, many lines of calculations are made to manipulate 4-vectors. These could be shortened and made less error prone by using TLorentzVector methods like, for example, Eta() and Px()

Let's converge today on which cleanup is quickly achievable so we can proceed.

sethzenz:
Ok that sounds fair enough.  A new PR, using a new rebased branch, is the right way to do it.  I'll look forward to the new PR.

PR #522, opened 2016-01-29
electron id scale factors

martinamalberti:
Electron ID scale factors for cutbased Id, loose working point.
Preliminary values from M. Sani: /afs/cern.ch/work/a/arun/public/forEGM/CutBased_textFiles_withSyst/CutBasedID_LooseWP_fromTemplates_withSyst_v1.txt

### DIFF
--- a/Systematics/python/flashggElectronSystematics_cfi.py
+++ b/Systematics/python/flashggElectronSystematics_cfi.py
@@ -1,11 +1,58 @@
 import FWCore.ParameterSet.Config as cms
 
 binInfo = cms.PSet(
-		variables = cms.vstring("pt"),
+		variables = cms.vstring("eta","pt"),
 		bins = cms.VPSet(
-			cms.PSet(lowBounds = cms.vdouble(1.0), upBounds = cms.vdouble(25.0), values = cms.vdouble(1.), uncertainties = cms.vdouble(.05,0.3)),
-			cms.PSet(lowBounds = cms.vdouble(25.0), upBounds = cms.vdouble(100.0), values = cms.vdouble(1.), uncertainties = cms.vdouble(.05,0.3)),
-			cms.PSet(lowBounds = cms.vdouble(100.0), upBounds = cms.vdouble(999.0), values = cms.vdouble(1.), uncertainties = cms.vdouble(.05,0.3))
+                        # CutBasedId - LooseWP scale factors
+                        # preliminary numbers from M. Sani: /afs/cern.ch/work/a/arun/public/forEGM/CutBased_textFiles_withSyst/CutBasedID_LooseWP_fromTemplates_withSyst_v1.txt
+
+                        cms.PSet( lowBounds = cms.vdouble( -5.000, 0.000 ) , upBounds = cms.vdouble(  5.000, 20.000  ) , values = cms.vdouble( 1.000 ) , uncertainties = cms.vdouble( 0.030 ) ),
+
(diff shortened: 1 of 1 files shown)


sethzenz:
@matteosan1 @martinamalberti I have a question about both these and the muon scale factors.   Are they intended to be linearly interpolated between bin centers, as is currently been implemented because it was done that way for leptons for Run 1?   Or are they intended to be scales which are applied in the same way for any location in a given bin (as is the case for photon scales)?

matteosan1:
I think that in pT the linear interpolation makes sense vs eta not much.

Regards,
     Matteo.

+++++++++++
Matteo Sani
University of California, San Diego, US.
Tel: +41227671577
CERN Office: 40-3-A02

sethzenz:
@molmedon can confirm, but I think the interpolation code isn't set up to interpolate in some variables and not others.  In fact I am not quite sure what it will do if there's >1 variable at a time:

https://github.com/cms-analysis/flashgg/blob/master/Systematics/interface/ObjectSystMethodBinnedByFunctor.h#L49-L91
https://github.com/cms-analysis/flashgg/blob/master/Systematics/interface/ObjectEffScale.h#L39-L114

Therefore the proposal is that we turn off all interpolation for the current round and revisit it later.  This does not require changing this PR; I'll make the change afterward.  But comments are appreciated!

matteosan1:
I think it make sense to remove the interpolation.

Regards,
     Matteo.

+++++++++++
Matteo Sani
University of California, San Diego, US.
Tel: +41227671577
CERN Office: 40-3-A02

ferriff:
OK with me.

molmedon:
Hi all, it is templated to take any kind of variable, but the algorithm is designed to handle 1d cases only. So for example Martina’s list contains two variables “pt” and “eta” the question would be to either interpolate in eta or pt ? If “pt” only, the python file should only have “pt” in the variables parameter, and each vector should only have one entry. If needed I can add the functionality to handle a 2d interpolation. But otherwise if only “pt” is used it should definitely work and I promise that it won’t crash any jobs.

Let me know what you think.

-Manuel



On Jan 29, 2016, at 7:31 AM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #522<https://github.com/cms-analysis/flashgg/pull/522>.

sethzenz:
The problem is then the interpolated curves can't be binned in eta, which is the form the SFs are given in.  Leaving out the interpolation for now is safer.  Later we can discuss whether any given set of SFs is intended to be used with interpolation or not, and if there is a need we can discuss how to implement interpolation with 1 interpoalted and 1 "spectator" dimension.

molmedon:
Also, the code expects both up/down uncertainties, so that’s why it was written with two values before as an example.  uncertainties = cms.vdouble( 0.030, 0.0 )

This would change if the 2d implantation is the way to go.

On Jan 29, 2016, at 2:07 PM, Manuel Alejandro Olmedo Negrete <manuel.alejandro.olmedo.negrete@cern.ch<mailto:manuel.alejandro.olmedo.negrete@cern.ch>> wrote:

Hi all, it is templated to take any kind of variable, but the algorithm is designed to handle 1d cases only. So for example Martina’s list contains two variables “pt” and “eta” the question would be to either interpolate in eta or pt ? If “pt” only, the python file should only have “pt” in the variables parameter, and each vector should only have one entry. If needed I can add the functionality to handle a 2d interpolation. But otherwise if only “pt” is used it should definitely work and I promise that it won’t crash any jobs.

Let me know what you think.

-Manuel



On Jan 29, 2016, at 7:31 AM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #522<https://github.com/cms-analysis/flashgg/pull/522>.

molmedon:
ok, this is ok with me.

On Jan 29, 2016, at 2:10 PM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:

any

PR #524, opened 2016-01-30
dijetMVA+combinedMVA weights update

yhaddad:
This PR includes the weights of latest VBF MVA retraining:
- Dijet MVA + combined MVA + VBF boundaries (VBF-1 & VBF-2)


sethzenz:
Looking at the plots, the boundary cuts are very tight, but I do see some events in both categories in signal MC.  So I take these for now and run workspaces, we have a look at the output, and then have another look at the optimization ASAP.

PR #533, opened 2016-02-07
Btagscalefactors

saghosh:
Including B Tag scale factors in FLASHgg


sethzenz:
Consolidated in #542 which supersedes this.

PR #537, opened 2016-02-10
Move some systematics manipulation into common python

sethzenz:
I have verified that the new workspaceStd.py perfectly mimics MicroAODToWorkspace.py, but a lot of its tricky manipulation is moved into separated python functions in Systematics/python/SystematicsCustomize.py

This will make it easier to see what is being done and find bugs.  N.B. I know of at least one bug (sigE/E smearing central value must be applied to data as well as MC) that will be fixed in a separate PR.

In the future MicroAODToWorkspace.py will be depreciated.  New/updated dumpers that need the final versions of objects can hopefully use the common functions.

### DIFF
--- a/Systematics/interface/BaseSystMethod.h
+++ b/Systematics/interface/BaseSystMethod.h
@@ -22,7 +22,8 @@ namespace flashgg {
             _Name( conf.getParameter<std::string>( "MethodName" ) ),
             _Label( conf.getParameter<std::string>( "Label" ) ),
             _MakesWeight( false ),
-            _RandomEngine( nullptr )
+            _RandomEngine( nullptr ),
+            _ApplyCentralValue( conf.getParameter<bool>( "ApplyCentralValue" ) )
         { }
         BaseSystMethod() {};
         virtual ~BaseSystMethod() {};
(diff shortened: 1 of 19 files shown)


sethzenz:
The third commit adds a major new feature, which based on recent experience is important to make sure that the workspaces are bug-free and that all plotting dumpers can be easily kept consistent.  It enforces a consistent ApplyCentralValue member in all systematics producers, allowing us to do an info dump of what each systematics producer is doing in a given configuration.  The results for signal, background, and data are shown here:

https://sethzenz.web.cern.ch/sethzenz/cms/SystList-Feb10.pdf

The most immediate issue I see is that sigEOverE isn't handled correctly:

* The central value is also misssing from data (my misunderstanding)
* It has syst shifts of +/- 0, which will be ignored. @vtavolar, I understand the expected behavior is +/- 1, in MC, which we should be picking up as a weight and propagating, right?

I could use some additional eyes (e.g. @musella, @ferriff, @martinamalberti ) on both this PR and the results in the PDF above, in particular:

* What object-driven systematics (so, excluding theory weights) should we have that aren't listed?
* What behaviors appear to be wrong?
* In each systematic producer, I have had to define the modification made when applyCentralWeight() is false.  This needs to be correct; sometimes it should be 0, sometimes 1, and sometimes (e.g. with smearing) there's no meaningful way to apply the systematic shifts unless the central value is also applied.  Each SystMethod's change needs to be checked to make sure I haven't put in a stupid default behavior.

Thanks!

sethzenz:
I think I may be getting mixed up between SigmaEOverEShift and SigmaEOverESmearing.  In any case, the PDF table says what the configuration does right now, so you can tell me how to fix it. :)

As a side note:  the default configuration in the cfi files is for signal MC;  modifications are made for the other two cases by the new customization methods.

PR #540, opened 2016-02-13
Photon ID in 76X

ferriff:
On behalf of @mmachet (please check the standard microAOD producer to be working properly).


sethzenz:
Runs fine.  Will merge.  Someone should eventually check that the output looks right in 76x.

PR #544, opened 2016-02-14
consolidated features for jets: studies, silver json func

sethzenz:
Selected work by @yhaddad @edjtscott and myself.  Original history can be seen here:

https://github.com/edjtscott/flashgg/tree/Ed-ZPlusJetTag-2

Features include:

* isGold method for tags
   * Needed for studies
   *  Could be used for excluding non-gold HF jets from analysis by "turning off" the VBF tag, if this is required
   * NB Not implemented in all tags yet - check!
* ZPlusJetsTag for studies only
* Additional VBF variables and dumping features for studies

### DIFF
--- a/DataFormats/interface/DiPhotonTagBase.h
+++ b/DataFormats/interface/DiPhotonTagBase.h
@@ -29,6 +29,8 @@ namespace flashgg {
         void setSystLabel( const std::string label ) { systLabel_ = label; }
         std::string systLabel() const { return systLabel_; }
         bool hasSyst( const string &label ) const { return ( systLabel_ == label );}
+        void setIsGold ( int runNumber );
+        bool isGold() const { return isGold_; }
     private:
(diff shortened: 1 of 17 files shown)


edjtscott:
Hi Seth,

Thanks for copying me into this. Are you able to meet tomorrow morning to discuss the status of the work below and what I should do next?

Also I am due to send an two or three line abstract to Björn for the 10 minute talk on the 25th Feb. Do you think the attached is suitable and do you have any suggestions for changes?

sethzenz:
@yhaddad You may be interested in looking at Systematics/test/ZPlusJetDumper_cfg.py for putting together (or comparing with) your updates to the VBFTagDumper.

yhaddad:
@sethzenz  Yes, I will add this tagger alongside the VBF dumbed tree.

PR #549, opened 2016-02-16
update-RV-scale-factors

malcles:


malcles:
Hello Seth,
please find updated scale factors for RVbins, using 76X Zmumu all stat + 76X Hgg MC all modes, final selection tagged events only. 
Julie

PR #557, opened 2016-02-18
Topic jec and interface update

sethzenz:
Update JEC interface because a new one was provided for CMSSW 76X.  The nicest way to accomodate this was to update the interfaces for the SystMethods, so this has been done.   The JetSystematicProducer no longer has unique code for setting JECs; instead, all SystMethods can be initialized based on the Event and EventSetup and have a consumesCollector() passed to them to load needed collections.  JEC is currently the only user, but this interface may prove useful in other cases in the future (e.g. loading rho for recomputing MVAs). 

Incidentally I fixed sigmaEOverE as discussed a week ago.

Tested for workspace dumping for signal, background, and data.


sethzenz:
Superseded by #566

PR #558, opened 2016-02-18
fix TnP version and cherry pick hash for pdf weight tools

sethzenz:
No promises but the changes here may make the code set itself up smoother.  Please watch for tags and hashes in setup.sh that don't work, and report them!

This may also help:
export CMSSW_GIT_REFERENCE=/cvmfs/cms.cern.ch/cmssw.git.daily


sethzenz:
Oops, I waited too long to merge and others did it!

PR #564, opened 2016-02-20
Run dependent scales and smearings

ferriff:
Posting the PR but needs testing prior merging. Recipe:
```
   cmsrel CMSSW_7_6_3_patch2
   cd CMSSW_7_6_3_patch2/src
   cmsenv && git cms-init
   git clone https://github.com/cms-analysis/flashgg flashgg
   cd flashgg
   git apply /afs/cern.ch/user/f/ferriff/public/setup.patch
   source setup.sh
```
At this point one gets complaints about some cherry-picks. Probably this is unrelated to the small change in `setup.sh` introduced by this PR, and it seems harmless.

The configuration of the systematics `Systematics/python/flashggDiPhotonSystematics_cfi.py` should already be working, but please check that the parameters are correct.


sethzenz:
Interface changes are slightly different than in #557, which is 100% my fault for not coordinating better when I knew you were working on this!  In any case it will be a quick fix to synchronize and test the two together.  I will merge and test it all tomorrow (along with nearly-finished JEC updates).

musella:
I will try to test this, but I think that something is not quite right in the setup and in the logic:
- I see no way to differentiate energy scale corrections and uncertainties for data and MC.
- The support for stocastic smearing seems to have disappeared? Is it delegated to the EGM class?

I have a very naive question: is there a reason why we did not simply extend the current binning adding the run number dimension?

sethzenz:
@musella The stochastic smearing is computed, for both "rho" and "phi", by the tool here:

https://github.com/cms-analysis/flashgg/pull/564/files#diff-4c399883ebac57db0915e3aaf9de38c9R70

ferriff:
@musella : 
   - I don't understand why the current logic is different from the previous implementation of 2D StochasticSmearing, but I am happy to modify/correct things
   - the reason not to extend the current binning is to move to an officially supported EGM tool that avoid us ad-hoc solutions of parsing and pasting O(100) lines in a cfg file (the treatment of SigmaEOverESmearing and SigmaEOverEShift is still suboptimal in this respect). Moreover, we are now transparent to possible changes in the parametrization.

musella:
Hi,


>    - I don't understand why the current logic is different from the
>    previous implementation of 2D StochasticSmearing, but I am happy to
>    modify/correct things
>
>
As far as I understand from reading the tool code, with the current
implementation the energy scale uncertainty would not be applied to MC,
because the tool always look up the uncertainties by run number (and the
config files are for data).


>
>    - the reason not to extend the current binning is to move to an
>    officially supported EGM tool that avoid us ad-hoc solutions of parsing and
>    pasting O(100) lines in a cfg file (the treatment of SigmaEOverESmearing
>    and SigmaEOverEShift is still suboptimal in this respect). Moreover, we are
>    now transparent to possible changes in the parametrization.
>
> I don't see why one would have to paste 100 lines in a cfg file, when the
configuration can be generated by a script.

In fact the systematics cff would probably benefit from a split, but that
is another story.

Concerning the officially supported EGM tool, I have the impression that it
is rather new and not debugged.
So even if switching to the EGM class for energy scale corrections (which I
personally would have not done three week before the target conference), I
don't see the reason to completely erase the existing (and tested)
implementation.

Cheers,
Pasquale






>
>
> —
> <https://github.com/cms-analysis/flashgg/pull/564#issuecomment-186565101>.
>



-- 
Pasquale Musella - ETHZ

ferriff:
Hi Pasquale,

   - how could the usual implementation look for non-run number dependent uncertainty for MC (bins are bins even for the old one)? As I said, I am happy to modify what is implemented. If it is just a matter of data/MC differentiation, isn't it enough to change the input file parameter (as one could have changed bins)?

   - the EGM tool is new in the EgammaAnalysis package, but the class is the one used in ECALELF since Run1. As per the switch: I decided to keep the same files and change their content to avoid proliferation of equivalent methods, but the magic of the git history allows to retrieve them all back as if never modified. BTW, the master is still unmodified in this respect. I am happy to duplicate, if there is consensus, just let me know.

How is the test going?

Cheers,
Federico

sethzenz:
@ferriff Why do you have items in your recipe (in the patch) that aren't in setup.sh?  If it's needed to run the code, it needs to be in the default branch after the pull request. Am I missing something?

sethzenz:
@ferriff In fact what this PR adds to setup.sh is a different branch than the ones in your patch:

ferriff:smearer_76X_test_v2

How does this one relate to the ones in the patch?  If I follow your instructions as written this branch will be ignored!

sethzenz:
FYI: I am proceeding assuming I should use the patch and ignore the changes made to setup.sh by the branch itself.   I do not see any specific or unusual complaints about cherry picks, although I do see several instances of "addinfo_cache" complaints.  These happen a lot and never seem to cause any trouble.

ferriff:
Hi Seth, yes please. I did not realize that branch went in the setup.sh
in the PR, it was an early test (which does not work properly).
The best is to use the patch as you did, I can fix the PR tonight if you
won't meanwhile.
Thanks!
Federico

sethzenz:
@ferriff OK, thanks for the confirmation.  I am going to play with integrating this PR this afternoon, we can talk tonight about where things are and what fixes are needed.

musella:
As we discussed in a private thread, I implemented the run-dependent corrections extending the flashgg binning features in #565.

With that we have a validated reference in flashgg including run-dependent corrections which does not touch the already validated scale and resolution systematic uncertainties.

I would thus suggest to rename the correctors based on the EGM tool to allow to use both.

sethzenz:
@musella @ferriff I will do a full integration of both implementations tomorrow, as you suggest with the name changed so that either can be run.

ferriff:
@musella : shall we still wait for your validation of my PR, as per previous agreement, or you decided to abandon it?

musella:
Hi,

I did not abandon the idea of validating the corrections done with the EGM tool.

On the other hand, my time budget to spend on this task is over for the moment. I hope I can have a look again next week.

In any case, if the new corrections are not disruptive, I guess that the PR go in w/o physics validation. 

Cheers,
Pasquale

sethzenz:
Partially superseded by #566, but python configurations are not included yet.  We still need to converge on the correct configuration for the final workspaces

PR #571, opened 2016-02-23
New diphoton MVA weights

vtavolar:
Hi,

this PR contains the latest weight file for the diphoton MVA, obtained under 76x.
The photonID mva used is computed using transformed r9,s4 and etaWidth in EB with a consistent set of weights.

The untagged categories boundaries are also updated.


sethzenz:
@vtavolar Can you comment briefly on why classes_def.xml for Photon needed updating when this PR doesn't change Photon?    Is this something that was missing from previous PR's (and if so do you know why they seem to run)?

vtavolar:
Hi Seth, 
I think this is actually not needed. The path to this PR was not perfectly linear and at some point I must have introduced a version 11 of the class. When rebasing to the master, the changes to the photon have been reverted and I had to increase the class version further to 12. I have just checked and the code compiles and runs (dumpers) with the classes_def.xml that is in the master (class version 10). I am going to revert the commit.

PR #576, opened 2016-02-24
Updating PhotonJetProducer to work in CMSSW_7_6_X

saghosh:
used for validaton using Gamma+jet events.
Values for Photon ID cuts updated, other miscellaneous changes made to have the code working in 76X


sethzenz:
@saghosh Does this need to be in the master for studies toward the Moriond analysis?  If not I would propose to hold it until things stabilize (sometimes next week).  Is that ok?

saghosh:
@sethzenz Yes, it is perfectly fine to hold it until later. I noticed while doing the validation that the master code was not updated, hence I made the changes. I have the modified code with me and as far as I am aware, I am the only one using this code as of now, so there should not be any problem with waiting until later to merge this.

sethzenz:
@saghosh What is the distinct meaning of vertex_index_ and the new vertex_index_jet_ in this object?  It looks like only the new one is used, but both variables exist.

saghosh:
@sethzenz The vertex_index_ is supposed to be associated with the vertex that is selected by the vertex id MVA and the vertex_index_jet_ is supposed to be for the vertex that is tagged by the jet. Currently the vertex_index_jet_ is taken as the 0th vertex since we use PFCHS jets, but it was put in case we want to loop over the vertices and select the vertex with the smallest dZ from the jet (or the highest pT charged track from the jet constituents).

PR #586, opened 2016-02-26
require vbf preselection to make tag

sethzenz:


sethzenz:
Untagged category too

PR #589, opened 2016-02-27
Add linear systematic to IDMVA

matteosan1:
Add linearly increasing systematic to IDMVA to cover low score region.


sethzenz:
Runs correctly with one tweak to workspaceStd.py that I will submit separately in a moment.

PR #595, opened 2016-02-28
too many now, kludgy, but reliable.  could be combined

sethzenz:


sethzenz:
N.B. wh and zh can't be put in the same directory -- the output file names are the same, because the inputs are.  Unless you have a fix specifically for that.

PR #596, opened 2016-02-28
Fix products of systematic weights, details in code comments

sethzenz:
Found a bug associated with how we treat weights for different object systematics when multiple objects are imported.  It only affected the TTH tags because only they use weights for multiple types of objects.  

The logic has been checked very carefully on an event-by-event basis.  Please see the code changes themselves for a detailed explanation, and for those who want to understand further, please consult the debugging code, an illustrative snippet of the output of which is now: https://sethzenz.web.cern.ch/sethzenz/cms/illustration3.txt


sethzenz:
N.B. No impact on central values.  But this did cause some object weight systematics to have both up and down "go in the same direction".

PR #602, opened 2016-03-06
Fix 2D systematics customization for non-signal MC

ferriff:


sethzenz:
Looks straightforward and correct, but I propose to hold it for post-Moriond because it's non-essential for any workspace reruns.  @ferriff Let me know if merging would make your life easier.

ferriff:
Never mind, thanks Seth. Post Moriond is fine with me. I merged with my
local branch meanwhile.

PR #611, opened 2016-04-06
Store other tag info

sethzenz:
* Store info on other tags interpretations: tag type (enum), category, and diphoton index
* Rewrite of TagSorter for clarity of selection steps, storage of info, and full debug info

Example of output run on TTH events, including readback test: https://sethzenz.web.cern.ch/sethzenz/cms/tagtest003.txt

@ferriff Can you take a look and let me know if this has the features you had in mind?


sethzenz:
Replaced by #619

PR #618, opened 2016-04-12
updated PDFweight producer

martinamalberti:
Rearrangements to use  xml parser in  PDF weight producer


sethzenz:
Tested for non-empty output against both powheg and mcatnlo samples, so different PDF samples are interpreted correctly.

PR #620, opened 2016-04-15
Topic ed jet validation update

edjtscott:
The JetValidationTreeProducer.cc now has additional branches for weights (just for Heppi plotting convenience, always set=1) and passesRMSIndex, to aid comparison with the VBFTag. The jetIndex has been altered to not count photons and there is now a boolean controlling whether the photon matching is done in the previous way or in the same way as the VBFTag. The config file you wrote allowing the VBFTag and the JetValidation to run over the same events is also included


sethzenz:
Tested.  Seems to work.  Thanks!

PR #622, opened 2016-04-22
creating the puweight output branch in GlobalVariablesDumper also for data

andreh7:
This allows merging of dumped trees from MC and data with hadd (otherwise, data trees which do not have the puweight branch are ignored by hadd). Setting puweight to one for data in GlobalVariablesComputer.

### DIFF
--- a/MicroAOD/src/GlobalVariablesComputer.cc
+++ b/MicroAOD/src/GlobalVariablesComputer.cc
@@ -149,9 +149,7 @@ namespace flashgg {
             cache_.nvtx = vertices->size();
         }
 
-        if (evt.isRealData()) {
-            cache_.puweight = 1.;
-        }
+        cache_.puweight = 1.;
 
         if( ! evt.isRealData() && getPu_ ) {
             double truePu=0., obsPu=0.;
--- a/MicroAOD/src/GlobalVariablesComputer.cc
+++ b/MicroAOD/src/GlobalVariablesComputer.cc
@@ -148,7 +148,11 @@ namespace flashgg {
         if( doVtx_ ) {
             cache_.nvtx = vertices->size();
         }
-        
+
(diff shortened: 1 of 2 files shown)


sethzenz:
@andreh7 What happens if the pu reweight is turned off for MC?  In this case it looks like only the data will have this branch and again there will be hadd issues..!   (On the other hand maybe this is not a case we should support at all..)

andreh7:
good point... the branch is created unconditionally in the patched version ( https://github.com/andreh7/flashgg/blob/2016-04-22-puweight-var/Taggers/src/GlobalVariablesDumper.cc#L83 ) so there should be no problem at the level of hadd.

Admittedly this is wasting a bit of space in the output tree when pileup reweighting is disabled. 

On the other hand, it's flashgg/MetaData/python/JobConfig.py which enables pileup reweighting (if puTarget is specified) except for data, so GlobalVariablesDumper currently can't not know if pileup reweighting is disabled in its configuration because the job is running on data or because it was disabled for all jobs.

One could modify JobConfig.py set the parameter 'puReweight' to True also for data and pass the pileup weights as it if were MC. The patch already sets the output variable puweight to one in GlobalVariablesComputer.cc if evt.isRealData() is true, independent of whether pileup reweighting is enabled or not.

The only other place where the parameter 'puReWeight' appears is in Validation/plugins/globelikeTreeMakerWithTagSorter.cc but this ignores pileup reweighting for data already and creates the branch whether pileup reweighting is enabled or not.

The easiest solution would be to create the puweight branch (in GlobalVariablesDumper) whether pileup reweighting is enabled or not.

However I just noticed that for MC and pileup reweighting disabled, cache_.puweight is not correctly initialized in GlobalVariablesComputer.cc, I can update the pull request.

PR #630, opened 2016-05-17
Topic MET corrections 80X 

lsoffi:
This branch can be used to run flashgg including the MET correction and uncertainties tool in 80X.
It should be ok to be run in 76X.
The following files have been changed:

In **setup.sh**:

-> 1. Add additional CMSSW packages
     git cms-addpkg PhysicsTools/PatUtils
     git cms-addpkg PhysicsTools/PatAlgos

-> 2.Change these two lines:
    https://github.com/cms-sw/cmssw/blob/CMSSW_7_4_X/PhysicsTools/PatUtils/python/tool
    /runMETCorrectionsAndUncertainties.py#L722-L723 to: 
             #jetCorrInputFileName = cms.FileInPath(jetUncInfos["jecUncFile"] ), #jecUncertaintyFile),
               jetCorrUncertaintyTag = cms.string("Uncertainty"), #jecUncertaintyTag),

In /MicroAOD/python/ add:
https://github.com/lsoffi/flashgg/blob/topic_metsyst/MicroAOD/python/**flashggMETs_cff.py**

In  MicroAOD/test/ add:
https://github.com/lsoffi/flashgg/blob/topic_metsyst/MicroAOD/test/**Fall15_25nsV2_DATA.db**
https://github.com/lsoffi/flashgg/blob/topic_metsyst/MicroAOD/test/**Fall15_25nsV2_MC.db**
https://github.com/lsoffi/flashgg/blob/topic_metsyst/MicroAOD/test/**microAODstdWithMETCorr.py**

In MetaData/work modify:
https://github.com/lsoffi/flashgg/blob/topic_metsyst/MetaData/work/**crabConfig_TEMPLATE.py**#L19
https://github.com/lsoffi/flashgg/blob/topic_metsyst/MetaData/work/**prepareCrabJobs.py**#L233

NB: The only parameter that has to be changed by hand,if one runs either on data or MC , for the time being is the "isMC" flag in the crabConfig.py produced.


sethzenz:
@lsoffi Why do we check out two packages in order to make one small change to a configuration file?  I looked at the code and I don't see any reason we can't run the existing python tool with some appropriate changes to the input (or modification of the output) in our own configuration files.  For everyone's information, the latest version of the methods in question are here:

https://github.com/cms-sw/cmssw/blob/CMSSW_8_0_X/PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py

mplaner:
I have to actually test in 808, but it looks like the changes to patutils are no longer needed.

The configuration now skips the line livia removed if the local JEC file isn't defined (so I think that's good).

https://github.com/cms-sw/cmssw/blob/CMSSW_7_4_X/PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py#L720-731

https://github.com/cms-sw/cmssw/blob/CMSSW_8_0_X/PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py#L768-788

For the JEC tag, it looks like it has been properly initialized in 80x to the value we want.  This wasn't the case in 74x.  Again, testing is needed to make sure I understand the behavior, but I suspect it will work out of the box in 80x:
https://github.com/cms-sw/cmssw/blob/CMSSW_8_0_X/PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py#L60

sethzenz:
@mplaner Thanks!  So as far as I understand, I am waiting for someone (you?) to test and provide a more minimal recipe. cc @lsoffi

mplaner:
Sure, I'll provide it.  It runs no problem in 808 without the modifications to cmssw.  

I'm also trying to put the db files in the /data directory and have crab access them from there, as then we don't need to modify the prepareCrabJobs or crabTemplate.  I haven't got that to work yet, the files are of course copied, but I haven't figured out what to have the python code look for to find the db files loaded from the /data directory.  I tried to follow the procedure in other files which access /data/files.

I'll add a pull request today, hopefully, a version which doesn't need to modify crab_TEMPLATE and prepareCrabJobs.py.

sethzenz:
I think this was superseded by #632 @lsoffi and @mplaner please let me know if anything is missing

PR #635, opened 2016-05-31
Tweaks from running high mass diphoton analysis in 80X

musella:

PR #640, opened 2016-06-07
updates to workspaces and for IC batch system

sethzenz:
1. Add ttHJetToGG_M*_13TeV_amcatnloFXFX_madspin_pythia8_v2 samples and update cross section line.  TO CONFIRM: that these samples are gg only.

2. Temporarily use GluGluHToGG_M-125_13TeV_powheg_pythia8.  TODO: CHANGE BACK WHEN AVAILABLE

3. increase allowed tag weight range to [0.5,2.25], because there was one TTHLeptonic event that returned 2.21

4. python handling of 80X puWeight objects, copied over from #635. 

5. IC-specific batch system hack, should not affect anyone else


sethzenz:
Updated the branch to fix the cross section syntax (apparently 1. is not a valid float in json, must be 1.0) and to use the right ggh_125 samples, which turn out to exist after all.

PR #646, opened 2016-06-10
Merge tweaks from high mass diphotons

musella:
* 80X PU
* masking of lumi-section for overlapping datasets (still to be tested on data)


sethzenz:
This is difficult to test in all cases as it is mostly python for various running circumstances.  However I don't see anything that will interfere with the normal workflows.  If anyone has a problem please contact me and I'll help sort it out.

PR #654, opened 2016-06-16
Mini-Isolation, TTH updates

saghosh:
Including mini-isolation in FLASHgg, cleaning the TTHLeptonic to remove double counting of jets for weights, using EGamma official recipeee for electron selection.


sethzenz:
Tested along with #655 to make sure they don't interfere.  Compiles ok.  MicroAOD runs ok.  I see TTH tags (hadronic, leptonic with electrons, leptonic with muons, leptonic with both (!)) in simple_tag_test.  All seems fine, will accept.

PR #660, opened 2016-06-23
New catalog for ICHEP (retrained MVAs)

ferriff:
Some datasets auxilliary to the analysis still to be finished, check the details with

   `fggManageSamples.py -C RunIISpring16DR80X-2_1_0-25ns_ICHEP16 list -v`

if you want the full name displayed.


sethzenz:
Question for @InnaKucher : do we need to update the xml files on these lines when we run ntuples/workspaces from these now microAOD?

https://github.com/cms-analysis/flashgg/blob/9b31cfb612e8c3a34c1518a2bf5e079eb21611a5/Taggers/python/flashggUpdatedIdMVADiPhotons_cfi.py#L10-L11

PR #661, opened 2016-06-23
added a parameter applyCorrections to DiPhotonWithUpdatedPhoIdMVAProducer

andreh7:
To be able to disable the corrections of the photon id MVA input variables for testing purposes.

The official version of the code tries to disable corrections when the file given with the 'correctionFile' parameter does not exist. However, this generates an exception:
```
python encountered the error: <type 'exceptions.RuntimeError'>
edm::FileInPath unable to find file
```

This patch allows to disable the corrections explicitly (the default setting of the new parameter is true in the corresponding _cfi.py file).


sethzenz:
@andreh7 - In the current version, if you comment out correctionFile entirely (rather than blanking the path as you seem to have) then things will work.  So I am not sure we need this change.

andreh7:
I didn't realize that the corrections can be disabled by commenting out the parameter altogether (I checked that it works for me).

To save some other people's time, please consider pull request #663 instead.

PR #666, opened 2016-06-29
ic-specific eos path and flag

sethzenz:


MetaData/python/JobConfig.py
```diff
@@ -345,7 +350,10 @@ def parse(self):
         if self.useAAA:
             self.filePrepend = "root://xrootd-cms.infn.it/"
         elif self.useEOS:
-            self.filePrepend = "root://eoscms//eos/cms"
+            if self.atIC:
+                self.filePrepend = "root://eoscms.cern.ch//eos/cms"
```
ferriff: Can't this full domain eoscms.cern.ch become the default also for people at CERN?

sethzenz: I vaguely recall that there are some sites where the existing formulation works and this one doesn't.  But I looked back at my discussion of the similar change to fggCheckFile.py and I can't find any mention of that.  Does anyone know?

PR #677, opened 2016-07-06
Pattern matching when reviewing catalogues and update to RunIISpring16DR80X-2_1_0-25ns_ICHEP16

vtavolar:
Hello,

with this PR is possible to match a pattern in the dataset name when reviewing a catalog using fggManageSamples.py

I used the new tool to merge the GluGluHToGG_M125_13TeV_amcatnloFXFX_pythia8 with its extension. This comes useful to avoid inadvertently picking also the same dataset with material variation in the catalog when running on muAODs. The merged catalogue is included in this PR.

Bests,
Vittorio

### DIFF
--- a/MetaData/python/samples_utils.py
+++ b/MetaData/python/samples_utils.py
@@ -403,13 +403,15 @@ class SamplesManager(object):
         if writeCatalog:
             self.writeCatalog(catalog)
 
-    def reviewCatalog(self):
+    def reviewCatalog(self, pattern=None):
         datasets,catalog = self.getAllDatasets()
 
         primaries = {}
         keepAll = False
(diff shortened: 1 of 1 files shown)


ferriff:
Hi, this is nice (I did it reviewing the whole catalog previously).

I suggest to accept the feature but leave untouched the catalog, as in
local I have merged it already with some more finished samples and it
would certainly conflict with this PR, which means I should re-import
and re-check everything. I can do the merging of the _ext samples tonight.

Federico

vtavolar:
Hi Federico, 
thanks for checking this PR. I was not aware of the conflict, now I reverted the commit merging the dataset.
Bests,
Vittorio

PR #682, opened 2016-07-08
photon id weights for ICHEP vertex included

InnaKucher:


sethzenz:
If I understand correctly, this feature didn't make it into 2_2_0.  So we should also update https://github.com/cms-analysis/flashgg/blob/master/Taggers/python/flashggUpdatedIdMVADiPhotons_cfi.py in order to use the latest ID MVA during the tagging/systematics step.  Can someone confirm? @InnaKucher @ferriff

ferriff:
Hi @sethzenz , for 2_2_0 I took the master version that you green-lighted to me, so I imagine this PR was not in.

InnaKucher:
Also we will update photon id again next week with data/MC corrections for input variables.

PR #687, opened 2016-07-11
recognize new miniAODv2 for pu distribution import

sethzenz:
The new uAOD sometimes don't have the expected string to identify which PU distribution to import.  This seems to fix the situation at least for the ntuples I've been looking at.  I checked in CMSSW and it doesn't appear any new PU distributions have been added that I should use instead:

https://github.com/cms-sw/cmssw/tree/0397259dd747cee94b68928f17976224c037057a/SimGeneral/MixingModule/python

However I would be grateful if someone can check and confirm.


sethzenz:
fix now seems ok

PR #693, opened 2016-07-12
Ad hoc reduction in output filename length

sethzenz:
This will hopefully fix the problem where output file length from the dumpers goes beyond 255 characters and outputs can't be created.  Totally untested, please let me know if it helps!


sethzenz:
Not needed, the outFile argument can be specified, something like this:

```
"GluGluMaterialUp" : [
["/GluGluVeryLongDatasetPath/WithManyCharacters/USER", { "args" :
["outputFile=output_GluGluMaterialUp.root"] } ]   ]
```

See mailing list discussion.

PR #709, opened 2016-07-14
Mass resolution decorrelation, tags and workspace configuration for fiducial xsec measurement

vtavolar:
Hi all,
this PR includes:
- the machinery needed to decorrelate the mass resolution, calculated on the fly for every event;
- additional tags and producers to run sigmaM categorisation instead of the usual tags;
- changes in workspaceStd.py so that from one python both versions of the workspaces (standard analysis and fiducial measurement) can be run. The default is the standard analysis, the fiducial one can be turned on specifying doFiducial=True on the command line.

I checked that the workspaces produced with this PR and with the master are perfectly synchronised on a sub-sample of signal events.

Since the PR makes the developments performed for the fiducial analysis converge into the master after quite some time, I think it would be good if someone could double-check it before merging.

Let me know if you have comments or suggestions,

Vittorio


sethzenz:
Hi @vtavolar - overall I like how you've organized this.  Keeping the flashgg master so it supports multiple measurements and analysis is something we'll need more of in the future, and this is a good model.  Some questions/comments

* Is Systematics/python/escales/Golden22June.py used anywhere?
* There is probably some common code for switching to the fiducial in workspaceStd.py (and other configs) that could be moved to Systematics/python/SystematicsCustomize.py ... but I don't think this is needed for now, just a note for the future.

vtavolar:
Hi Seth,
thanks for having a look at this. About your comments:
- the file is not used anywhere for the workspace production, I'll remove it from the PR
- I agree on that, I guess for instance the manipulation of  the tagsequence and the tagsorter would better belong to the systematics customization config. As you said, we can keep it as it is for the moment and organise it better in the future.

Thanks,
Vittorio

PR #726, opened 2016-07-18
Photon id retrained with input variables shifted

InnaKucher:


sethzenz:
Holding this until it's time for it to be integrated into the workspaces in a self-consistent way. cc @martinamalberti @ferriff

sethzenz:
Better to keep existing ID MVA as long as subsequent MVAs are expecting it

PR #743, opened 2016-07-24
Final shower shape correction file for ichep 2016

kmondal:
Shower shape correction file has been updated


sethzenz:
Copied over into #745

PR #753, opened 2016-08-25
re-added support for authentication when performing queries to DAS

### LINKED ISSUE #752: fggManageSamples.py needs das_client.py supporting authentication
Since a few days, DAS requires authentication to read the information (see e.g. cernCompAnnounce/1144/1.html on hypernews). 

This can be achieved (assuming the user has a valid grid proxy) by adding the keyword arguments

`ckey=x509(), cert = x509()
`

to the calls to `das_query(..)` in MetaData/python/samples_utils.py (I can produce a pull request if desired). In fact, some commented out lines have these arguments already.

However the copy of das_client.py we have in MetaData/python/das_client.py does not support such authentication yet. We could either just remove it and rely on the one provided by the CMSSW environment or copy the latest version into the flashgg repository. 

The latter would have the advantage that it would also work with older CMSSW versions such as CMSSW_8_0_8_patch1 which points to das_client v02.05.00 which seems not to support authentication.

andreh7:
needed since some time, e.g. to do `fggManageSamples.py import <dataset>`.

Should fix #752 . 

Checked the modifications -- when running without a grid proxy, one now gets an error message when running `fggManageSamples.py import <dataset>` also showing the command how to obtain a grid proxy.


sethzenz:
@andreh7 Thank you!

PR #770, opened 2016-10-31
clone json for next production; 2 muon events only

sethzenz:


sethzenz:
This PR reflects that single muon is not needed for ZMuMuGamma.  This is to be double-checked and will be discussed in the Hgg meeting today.  Of course the changes can be reversed if needed but I accept for now.

PR #776, opened 2016-11-07
added costhetastar variable to VHH class & producer

molmedon:


sethzenz:
Why can't this already be computed via a function that uses the calculation in the producer, but without adding a new member variable?

molmedon:
Not sure if I understood your comment. The idea is to simply be able to dump this variable into a tree. Along with mjj, the other main cut in vhh is costs. Mjj though can be calculated with the two passing jet info (later) while for costs we also need the diphoton.



On Nov 8, 2016, at 5:32 AM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


Why can't this already be computed via a function that uses the calculation in the producer, but without adding a new member variable?

sethzenz:
What's required to dump a variable into the tree is a simple member function in the Tag class.  That does not require that an extra member variable be used to store the information: the jets and the photons are all accessible through the tag already. A rough version of the needed function could look like:

float CosThetaStar() {
            TLorentzVector jetl, jets, dijet, diphoton, vstar;
            diphoton.SetPtEtaPhiE( diPhoton()->pt(), diPhoton()->eta(), diPhoton()->phi(), diPhoton()->energy());
            jetl.SetPtEtaPhiE( leadingJet()->pt(), leadingJet()->eta(), leadingJet()->phi(), leadingJet()->energy() );
            jets.SetPtEtaPhiE( subLeadingJet()->pt(), subLeadingJet()->eta(), subLeadingJet()->phi(), subLeadingJet()->energy() );

            dijet = jetl + jets;
            vstar = diphoton + dijet;

            diphoton.Boost( -vstar.BoostVector() );
            float costhetastar = -diphoton.CosTheta();
	return costhetastar;
}

I guess in the end an extra variable would be ok too; this function is a bit more complicated than I had been thinking.

molmedon:
aah, ok I see what you mean. Yes in principle you can just dump all the required info and recalculate later in a simple macro. It is just so much more convenient if could just directly be dumped. Does changing the class def affect thing the code upstream ?

On Nov 8, 2016, at 8:11 AM, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


What's required to dump a variable into the tree is a simple member function in the Tag class. That does not require that an extra member variable be used to store the information: the jets and the photons are all accessible through the tag already. A rough version of the needed function could look like:

float CosThetaStar() {
TLorentzVector jetl, jets, dijet, diphoton, vstar;
diphoton.SetPtEtaPhiE( diPhoton()->pt(), diPhoton()->eta(), diPhoton()->phi(), diPhoton()->energy());
jetl.SetPtEtaPhiE( leadingJet()->pt(), leadingJet()->eta(), leadingJet()->phi(), leadingJet()->energy() );
jets.SetPtEtaPhiE( subLeadingJet()->pt(), subLeadingJet()->eta(), subLeadingJet()->phi(), subLeadingJet()->energy() );

        dijet = jetl + jets;
        vstar = diphoton + dijet;

        diphoton.Boost( -vstar.BoostVector() );
        float costhetastar = -diphoton.CosTheta();
return costhetastar;


}

I guess in the end an extra variable would be ok too; this function is a bit more complicated than I had been thinking.

PR #779, opened 2016-11-11
setup updates

sethzenz:
To fix 2 bugs discussed on the flashgg list:

* Change in MET recipe breaks setup.sh
* ConfigToolBase.py needs to be tweaked or it breaks for default fggRunJobs.py config on lxbatch


mplaner:
Confirmed by MET group, this will take the proper JES corrections into account for MET.

PR #782, opened 2016-11-14
Gen variables, extra floats, scripts to reduce ws for differential analysis

vtavolar:
This PR merges the classes for gen-level analysis needed for the differential analysis.
More modifications to the extra floats handling are introduced, in order to allow the possibility to add variables to the workspaces as extra floats.

Few other modifications/improvements are contained in this PR:
- additional trees for gen-level analysis;
- optimised hadd_workspaces routine, with possibility of parallelization (gain a factor ~400 in wall time);
- scripts to reduce and split "inclusive" fiducial workspaces into "differential" workspaces.

I have produced the ws for the main analysis with the master branch and with this branch, and I have obtained the same results (same number of entries in the datasets).

Please let me know if you have comments/requests,
Vittorio


sethzenz:
Hi @vtavolar, sorry for getting to this only slowly, but here's a first comment:

Can you get rid of the changes to MetaData/data/RunIISpring16DR80X-2_2_0-25ns_ICHEP16_MiniAODv2_Training/datasets.json ?   They look accidental.

vtavolar:
Hi @sethzenz,

I am quite sure there was a reason for those changes back in July, in order to avoid failure of a few jobs. Anyway, since nobody experienced problems recently with the catalog, I have reverted them.

Bests,
Vittorio

sethzenz:
In principle I've now gone through the code and checked for potential clashes with the existing workflow, and nothing stands out.  I guess the next step will be some tests of the dumping to double-check that all is well.  Do you have any other suggestions for where code should be checked carefully?

In the meantime, I do have two outstanding comments:

* Some files still removed from MetaData/data/RunIISpring16DR80X-2_2_0-25ns_ICHEP16_MiniAODv2_Workspaces/datasets.json
* root://eoscms.cern.ch//eos/cms vs root://eoscms//eos/cms   — see flashgg list email

sethzenz:
@vtavolar After I merge this into the current upstream/master, workspaceStd.py crashes (e.g. with processId=tth_125) because customize.datasetName() returns NoneType -- so the "count" method cause an exception.  I can simply remove this check (which in baseline flashgg+ main analysis should never be set to anything useful, as far as I know) and continue testing.. but in the meantime, what do you think the fix should be?

PR #787, opened 2016-12-01
For new PU reweighting

JunquanTao:
Modified the JobConfig.py to add one additional option "puSIM" to taking the true MC simulated PU distribution (normalized to 1.0) for PU reweighting. When running the new PU reweighting with the input "puSIM" option, one need to specify the input puSIM values, with the similar usage as "puTarget". For "Spring16 80X" MC samples,  puSIM=0.00131931,0.00230159,0.00346567,0.00483143,0.00642775,0.00836145,0.0107535,0.0135679,0.016785,0.0203496,0.0241871,0.0282775,0.0324902,0.0366713,0.040648,0.0442128,0.0472559,0.0496434,0.0511215,0.0518062,0.0514804,0.0503577,0.0484196,0.0458103,0.0427031,0.0392474,0.0354694,0.0317312,0.0279578,0.0243083,0.0208245,0.0176106,0.0147456,0.0121483,0.0098674,0.0079237,0.00625031,0.00487828,0.00374769,0.00283541,0.0021124,0.00156391,0.00112964,0.00079719,0.000570062,0.000388339,0.000267111,0.000179967,0.000119864,7.76474e-05  (50 values) which were taken from GJet_Pt-40toInf MC sample with simulated events N~58M (almost the same values from DiPhotonJetsBox_MGG-80toInf with ~38M events). See details from slide 6 in ref [1].

Without this "puSIM"  option, the probvalues from OOT cfi file will be used for PU reweighting, which is the same as what we used for ICHEP16.   

[1] https://indico.cern.ch/event/566824/contributions/2390463/attachments/1380995/2099339/20161201IssuesOnSigFit_LowMass2016data_HggWorkingMeeting.pdf


sethzenz:
@JunquanTao Sorry that I'm a little slow to understand, but this PR is not changing as much as I expected.  The calculation method doesn't change at all; it's simply that the simulated PU event count is being replaced by a user-specified one.  So is the conclusion that the "OOT cfi file" is simply wrong?  Or were we using it the wrong way before?  We need to understand this. cc @ferriff @martinamalberti @musella

JunquanTao:
@sethzenz  In this way, others can also compare the validation results from the default one (taking the probValues from the "OOT cfi file" as input) and the user-specified one (taking the suggested puSIM values as input when running fggRunJobs.py ). The input puSIM values can be changed easily according to different simulation condition, such as from Spring16 to Moriond17.  

By default in the codes,  the probValues from the "OOT cfi file" were used as input for mcpu  (GlobalVariablesComputer.cc) to calculate the PU weights.  Then PU weight is assigned to each MC event according to the true Npu in this MC event in the codes. 

The codes for the calculation of PU weights (GlobalVariablesComputer.cc) will not be problem in technology, if the input mcpu values are reasonable.  As shown in my slide 5 [1] or checked from the "OOT cfi file" directly for Spring16,  the problem is that the probValues are 0 when Npu>38 until Npu up to 50. These 0 values in the denominator for the calculation of PU weights in  GlobalVariablesComputer.cc will  cause PU weight problem, as we have seen from our lower mass signal modeling.  

From my investigation with  ICHEP dataset and Spring16 80X DY MC, it's better to use the true simulated PU distributions as  input for the calculation of PU weights, when looking at the Nvtx distributions from Zee data and MC.

With this PR changes, one can also compare the validation results from two methods (probValues from the "OOT cfi file" or from the true simulated PU), with the Zee events from re-reco data and the exist Spring16 MC or  Moriond17 MC in near future.

[1] https://indico.cern.ch/event/566824/contributions/2390463/attachments/1380995/2099339/20161201IssuesOnSigFit_LowMass2016data_HggWorkingMeeting.pdf

sethzenz:
@JunquanTao I understand the change you've made in this PR from a technical perspective.  However, we need to draw a conceptual conclusion in order to move forward and make sure everyone draws the right recipe.  The old procedure and your procedure fill in values for puObj.mcPU that are qualtatively different from each other [1], but then they use the values in the same way in subsequent processing.  We need to understand which set of numbers is the right one, or whether our calculation procedure corresponds to one set of numbers and not the other.  I'll start a thread on the flashgg list...

[1] For everyone's reference, the two different value lists set for puObj.mcPU that are set by the script are below.  They definitely peak in different places.

fromTao = [0.00131931, 0.00230159, 0.00346567, 0.00483143, 0.00642775, 0.00836145, 0.0107535, 0.0135679, 0.016785, 0.0203496, 0.0241871, 0.0282775, 0.0324902,
0.0366713, 0.040648, 0.0442128, 0.0472559, 0.0496434, 0.0511215, 0.0518062, 0.0514804, 0.0503577, 0.0484196, 0.0458103, 0.0427031, 0.0392474, 0.0354694, 0.0317312,
0.0279578, 0.0243083, 0.0208245, 0.0176106, 0.0147456, 0.0121483, 0.0098674, 0.0079237, 0.00625031, 0.00487828, 0.00374769, 0.00283541, 0.0021124, 0.00156391, 0.00112964,
0.00079719, 0.000570062, 0.000388339, 0.000267111, 0.000179967, 0.000119864, 7.76474e-05]

fromCFI = [0.000829312873542, 0.00124276120498, 0.00339329181587, 0.00408224735376, 0.00383036590008,
    0.00659159288946, 0.00816022734493, 0.00943640833116, 0.0137777376066, 0.017059392038,
    0.0213193035468, 0.0247343174676, 0.0280848773878, 0.0323308476564, 0.0370394341409,
    0.0456917721191, 0.0558762890594, 0.0576956187107, 0.0625325287017, 0.0591603758776,
    0.0656650815128, 0.0678329011676, 0.0625142146389, 0.0548068448797, 0.0503893295063,
    0.040209818868, 0.0374446988111, 0.0299661572042, 0.0272024759921, 0.0219328403791,
    0.0179586571619, 0.0142926728247, 0.00839941654725, 0.00522366397213, 0.00224457976761,
    0.000779274977993, 0.000197066585944, 7.16031761328e-05, 0.0, 0.0,
    0.0, 0.0, 0.0, 0.0, 0.0,
    0.0, 0.0, 0.0, 0.0, 0.0]

sethzenz:
@JunquanTao Just to confirm, when you get your better-agreeing plots, you are using the weight output from ntuples coming out of the JobConfig.py code?  

I.e. you are not applying any extra weights or external calculations?  (For example, pileup weight is already included in the main weight and should not be applied again.)

mplaner:
@JunquanTao I'm confused by your statements.  From your presentation, I thought you were using reconstructed nPU from MC and reconstructed nPU from data to derive the weights.  But now it sounds like you are using the true nPU from MC.  But I don't understand how you can get a true nPU from MC greater than what's in the OOT cfi file for the generated MC.

Can you clarify exactly which variables were used for the inputs for your weights?

JunquanTao:
@sethzenz  I draw the Nvts plots from the dumpTrees of Zee data/MC and applied the "weight" with PU weights calculated using "fromTao" instead of "fromCFI" for input mcpu values.  The PU weights should be calculated in GlobalVariablesComputer.cc and controlled by JobConfig.py.  I don't think I have applied the PU weights twice.

@mplaner  Sorry for the confusion.  I was using the "simulated and reconstructed Npu" in MC (getPU_NumInteractions, or obsPU in the codes) and reconstructed nPU from data to calculate the weights. I did not use these Prob Values from CFI. 

 @sethzenz It's better not taking further action on this PR. Although the changes did not change the default way we used for PU reweighting in past. I need to understand the "weight" and the analysis flow of Zee validation more precisely.

MetaData/python/JobConfig.py
```diff
@@ -268,7 +278,11 @@ def customize(self,process):
...
                             puObj.puReWeight = True
                             puObj.puBins = cms.vdouble( map(float, samplepu.probFunctionVariable) )
-                            puObj.mcPu   = samplepu.probValue
+                            #Using input MC PU or by default the probvalues from cfi file for PU reweighting
+                            if pusimu:
+                                puObj.mcPu   =  cms.vdouble(pusimu)
+                            else:
+                                puObj.mcPu   = samplepu.probValue
                             puObj.dataPu = cms.vdouble(putarget)
                             puObj.useTruePu = cms.bool(True)
```
musella: As discussed on the egroup, this line need also to be changed to use the "obseved PU".

@sethzenz, looking at the code, I realized that the variable useTruePu is not initialized in C++ if not specified in python. 
 https://github.com/cms-analysis/flashgg/blob/master/MicroAOD/src/GlobalVariablesComputer.cc#L65
Given that the python always set the flag, we are safe, but it would be good to initialize the flag in the constructor.

sethzenz:
@JunquanTao Is this PR still intended to go in?  A lot of commits were just added to it.

JunquanTao:
@sethzenz  Most of us now are trying to use Moriond17 samples. We have all the prob. values with the true Npu from 0 to 75 and the reweighting method with true Npu works very well. This PR (#787) can be dropped. Thanks!

sethzenz:
OK, thank you!

PR #795, opened 2017-01-15
Topic ele egmid 15 jan reg

sethzenz:
Includes new regression recipe from EGM provided today.  Also additional photon IDs from EGM required for HH.  Testing and validation still required


sethzenz:
Expect this will not be accepted, because we decided to use the old regression.

sethzenz:
Accepting as a reference for the HH production, but with new regression turned off by default

PR #798, opened 2017-01-19
Photon id weights, photon isolation corrections and shower shape transformations

InnaKucher:
Hello,

please find new photon id weights, the code is adapted to compute proper photon isolation for EE and shower shape transformations are added.

Inna


sethzenz:
@InnaKucher Can you please have a look at running the MicroAOD step with this PR branch?  I ran some tests/debugging and I think it's crashing here:

https://github.com/InnaKucher/flashgg/blob/f8299c29feda37937c959402a348804188347025/MicroAOD/plugins/PhotonProducer.cc#L287

I suspect this is because the additional arguments have not been copied from the DiPhotonWithUpdatedPhoIdMVAProducer.cc

InnaKucher:
@sethzenz Hello Seth, yes, you are right. I forgot to add those changes to photon producer. This commit should fix it:
https://github.com/InnaKucher/flashgg/commit/d62bcbf34a6634b82b485bf21b83fcae9f31bc2c

I tested MictoAOD step with those changes.

Inna

sethzenz:
@InnaKucher Thanks for the quick action. Accepting now!

PR #794, opened 2017-01-11
Updates for TTH Hadronic MVA

saghosh:
Updates for ttH Hadronic MVA, 
new BTag WPs for Moriond2017, 
 modified method of multiplying object weights so as to have includeWeightsByLabel in WeightedObject.cc,
synchronised B Tag weights method with central cmssw tool

PR #800, opened 2017-01-21
Updates for summer 16 electron ids and their offline calculation for data

saghosh:
Updates for electron id synchronised with recommendations for post ICHEP16 at MicroAOD level for data and using the stored VID based IDs for MCs.

Also storing the summer 16 electron id MVA values


sethzenz:
Removed minor conflicts with #794 using the web interface.  Hopefully I got it right!

PR #803, opened 2017-01-23
tweak matching and truncation per latest JetResolution twiki Smearing…

sethzenz:
See https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetResolution#Smearing_procedures for the math that is being changed (slightly).  This only became an issue because there is now a bin with SF < 1 with the latest data and MC.

Thanks to @gkrintir for pointing out this issue


sethzenz:
The implementation of matching cuts as described on the TWiki (unrelated to the bug) increases the number of jets falling to the "unmatched" case by about 15%.  (Previously, the definition of "match" was simply the presence of matched genJet as defined in pat during MiniAOD.)  This seems reasonable.

PR #804, opened 2017-01-23
Split catalog into multiple file to avoid git file size issue

musella:
Catalogs are split automatically in such a way that the size of json files stays reasonable.

In practice the catalog is split in such a way that no file contains information about more than 1k root files. 
Datasets are never split, therefore the threshold on the number of files can be passed if the file contains just one dataset.

The dataset is split when the catalogs is written back to disk (after any command that potentially modifies the catalog, such as `import`, `review`, `check`, etc..) and a new command `rebuild` is added to force the splitting of existing catalogs.


sethzenz:
@musella Sorry for slowness of accepting.  This will be used as we catalogue the reMiniAOD-->uAOD output.

PR #813, opened 2017-01-27
Combined MVA for Moriond 17 + VBF boundaries

yhaddad:
Combined VBF MVA with the necessary + changes on the VBFMVA.


sethzenz:
Note: one small fix will come in a separate PR after this is accepted.

PR #817, opened 2017-01-31
Differentials: trees reduction to ws, global vars binning, CompositeCandidates in the tag

vtavolar:
Dear all,
this PR contains the status of the framework in view of the differential measurements for Moriond. It introduces the following features:
- "inclusive" trees are produced instead of workspaces, which are obtained at a later stage when splitting the events into the bins/processes of the analysis;
- the handling of the binning of the global variables has been improved, adding also the possibility to specify a variable binning;
- CompositeCandidate objects are added to the differential tag. This allows to attach, for instance, the jet objects to the tag, taking advantage of the tag replication for systematic variations. 

I would suggest not merge it just yet because I haven't had time to check that no variation is introduced in the standard workflow.

Bests,
Vittorio


sethzenz:
@vtavolar I guess we need to sort out a new clean version of this of this to integrate.  Is that still on your to-do list?

PR #826, opened 2017-02-03
Topic vbf boundaries 3 feb

sethzenz:


sethzenz:
Messy history but changes once flattened are:
* For workspaces, puTarget as agreed today (cms default)
* Use VBF boundaries as shown today

PR #831, opened 2017-02-07
removed merging of CMSSW_8_0_X-METFilterUpdate, using commit 7a07643 instead

andreh7:
(as discussed with @mplaner)

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -108,8 +108,11 @@ echo "Setting up QGL..."
 git cms-addpkg RecoJets/JetProducers
 git cms-merge-topic -u sethzenz:for-flashgg-QGL-vertexIndex-8_0_20
 
-echo "grabbing MET topic updates..."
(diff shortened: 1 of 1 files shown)


andreh7:
I just noticed that this for example will not checkout everything properly
(e.g. RecoMET/METFilters/python/BadChargedCandidateFilter_cfi.py is not checked out, this is needed for flashgg/MicroAOD/test/microAODstd.py).

Probably can be fixed by also updating .git/info/sparse-checkout by hand.

sethzenz:
@andreh7 Thanks for this useful example of how to manipulate sparse checkout. I believe in the end this is superseded by the update to #843 , but please let me know what you think

andreh7:
yes it's superseded (unless somebody like me needs to rerun a microAOD before re-miniAOD), so feel free to close it

PR #835, opened 2017-02-08
Pile up jet id systematics producer

yhaddad:
This PR contains a PUJID systematic producer and the related uncertainties on pt and eta bins.


sethzenz:
@yhaddad I believe this PR accidentally reverts recent changes to the btag weights.  Can you check please?

PR #839, opened 2017-02-09
Adding cMVA b-tagger

rateixei:
Storing extra b-tagger in jets.


sethzenz:
On hold after discussion with @rateixei .  Missing producer and insufficient time/urgency to investigate.

PR #851, opened 2017-02-11
STXS Dumping

sethzenz:
* Dumping code updated for HTXS (the hard part is splitting the compressed PDF weights).
* Customizable option for doing HTXS dumping instead of the regular scheme


sethzenz:
Superseded by #851

PR #867, opened 2017-02-20
Update flashggMuonSystematics_cfi.py

gkrintir:
ok @sethzenz @saghosh here the suggested change


sethzenz:
@gkrintir You've confirmed that this runs ok and fixes the exception you posted?

gkrintir:
Forgot to mention. Right, I confirm that.

If you agree we can extend it also to e.g. Electron.

sethzenz:
It seems like a good thing to do, for anywhere we have an arbitrarily high pt or energy bin.

PR #871, opened 2017-02-21
Scales and smearing corrections update

vtavolar:
Hi all,
in this PR I include in flashgg the new set of energy scale and smearing corrections taken from https://github.com/gfasanel/ScalesSmearings/commit/bf3fd0e30e129aaef377819aba3e9c15a9179182 , with the relevant changes to copy/read them in flashgg.

They are derived on full 2016 dataset, regression 74, for photons.

Bests,
Vittorio


sethzenz:
General note: the change in setup.sh means that the additional copy commands will have to be run.  I'll make a note of this in email I send about the next tag.

PR #873, opened 2017-02-21
apply iso correction only on real data

sethzenz:


sethzenz:
Code was right, PR title was exactly wrong.  Fixed for clarity.  Thanks @andreh7 for pointing this out

PR #874, opened 2017-02-21
Jet ID Updates

sethzenz:
Three things are done in this PR:
* Update JetID prescription to include changed recommendation for transition region, see https://twiki.cern.ch/twiki/bin/view/CMS/JetID13TeVRun2016 
* Remove passesPuJetId call in tags, which always returns true anyway.
* Consistently put in Loose Jet ID for all tags using jets, per JME recommendations.  You can see in https://indico.cern.ch/event/592313/contributions/2398387/attachments/1384663/2106630/JetID_JMARmeeting_7_12_2016.pdf that efficiency is >99% throughout, so no impact is expected on the analysis.   (The exception is VBF, which needs Tight Jet ID.)

### DIFF
--- a/DataFormats/src/Jet.cc
+++ b/DataFormats/src/Jet.cc
@@ -81,21 +81,24 @@ bool Jet::passesJetID( JetIDLevel level) const
     
     //std::cout  << "DEBUG:: eta= " << eta << " NHF=" << NHF << std::endl;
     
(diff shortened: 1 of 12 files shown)


sethzenz:
Impact quite small as expected.  Yield of data changes by under 1% for all tags, and for signal the yield for "diagonal" process,tag pairs [3] always changes by under 0.25%.  A few off-diagonal tag,proc pairs change by more, which is not surprising or a cause for concern, because these cases have relatively few events, and both positive- and negative-weight events can move.  A complete list of the yield changes is at [4].

Seth

[3] "Diagonal" means that the process is targeted by the tag, e.g. WH with W --> jets and VHHadronic.
[4] https://sethzenz.web.cern.ch/sethzenz/cms/JetID_yield_compare.txt

PR #882, opened 2017-02-27
Moriond 27 feb2017

ArnabPurohit:
Dumpers for TnP measurements of photons have been updated


sethzenz:
I can't vouch for the details, but the logic seems sensible and this won't have any impact on other analysis steps.

PR #886, opened 2017-03-03
Topic misc rebase2

sethzenz:


sethzenz:
Features in this PR:
* HTXS tools (can be turned on or off with flag)
* NoTag object for tracking QCD weights of signal not entering analysis
* CentralHiggsFilter for HTXS for signal samples that didn't have HTXS stored, e.g. bbH

PR #906, opened 2017-03-14
workspace sh updates

sethzenz:
Labeling, consistency, switch DoubleEG to new processing (with MET fix) to test.


sethzenz:
Also, move presently unused files in Systematics/test to Systematics/test/old.  (If any I moved are actually current, let me know and I'll reverse.)

PR #915, opened 2017-04-10
Replace ttH sample with correct config

sethzenz:
The ttH 126 sample had the wrong configuration.  This is fixed in this catalog change.     (Note there is still a catalog update needed for replaced material samples with the same issue, but I separate this PR because it's the only sample that affects the final workspaces.)


sethzenz:
Actually some of the fixed material samples are in there too but I still need to double check them.

PR #918, opened 2017-04-12
fixing basic workflow test

michelif:
Ciao,
today I've run flashgg for the first time. I've run the examples of flashgg and they do not run out of the box, since there are some errors. I fixed them, if you think is useful (to avoid the debugging to new users) merge the pull request.

I've changed three things:
= In the Readme there is "cmsRun MicroAOD/test/microAODstd.py", if you don't specify the options "processType=sig datasetName=glugluh" the second step crashes
= cmsRun Taggers/test/diphotonsDumper_cfg.py doesn't work because:
1. the global tag was missing, causing an error related to GBRDWrapperRcd class
2. the variable r9 has changed name (r9->"old_r9" or "full5x5_r9")

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ If everything now looks reasonable, you can build:
 And a very basic workflow test:
  ```
  cd $CMSSW_BASE/src/flashgg
- cmsRun MicroAOD/test/microAODstd.py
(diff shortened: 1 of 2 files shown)


sethzenz:
Looks good, thank you!

PR #925, opened 2017-06-07
Add b tag discriminant cmva v2

saghosh:
Preliminary updates to include the bTag discriminant CMVAv2 in flashgg along with CSVv2.

Rerun IVF as suggested by BTag experts for non default primary vertex.

Pull request done so that changes to the code are easily visible to others in the Hgg working group.


sethzenz:
Accepting this PR creates this warning in 92x:

> -------------------------------------------------------------------
>  Warning: You are attempting to remake IVF secondary vertices from
>           MiniAOD. If that was your intention, note that secondary
>           vertices remade from MiniAOD will have somewhat degraded
>           performance compared to those remade from RECO/AOD.
> -------------------------------------------------------------------

To be investigated.  Nevertheless we are only adding new information to the jets, so I'll accept the PR for now to move forward...

saghosh:
Hi Seth,

I think the plan was remake the secondary vertices, since we don't use the
default primary vertex in H->yy and we actually create the set of jets
corresponding to different primary vertices so the secondary vertices are
to be regenerated correspondingly.

With regards,
Saranya Ghosh

andreypz:
Hi,

For HH->bbgg analysis, the CMVA (and maybe other b-taggers?) would be quite useful, even with respect the default primary vertex. In our case 99% of  the times the default vertex is selected.

This is just to request to keep default b-taggers computed with respect default primary vertex.
Which maybe means to save a default jet collection? (I'm not sure if it's already saved or not).

saghosh:
Hi,

The values of the CSVv2 and the CMVAv2 will be stored also for the jet
collections of the zero-th primary vertex.

There is an option to not re-evaluate the secondary vertices, but then that
would be the case for all the jet collections in the way it is implemented.
The recommendation was to recompute the secondary vertices since for some
of the jet collections, the primary vertex is not the zero-th one.

If the flag is changed from true to false in:
https://github.com/cms-analysis/flashgg/pull/925/files#diff-2a98a870d06e8f73fb46887dee585251R210

Then the default set of secondary vertices will be used (that correspond
only to the zero-th primary vertex, even though the jet collection may be
made with a different primary vertex).

With regards,
Saranya Ghosh

sethzenz:
There's another wrinkle here, which is that by default we remake even the
0th vertex jets.  We should probably change this to save the old ones,
and/or  run a test where we save both the originals and the remade ones and
see how they compare.

andreypz:
Yes, I think that would be great to also save unchanged 0th vertex jets!

@saghosh do you know how often do you select 0th vertex versus non-0th in the ttH selection? I would guess it's like in bbgg - most of the time you would get the 0th.

PR #926, opened 2017-06-14
Allow compilation in CMSSW_9_2_0

sethzenz:
Numerous code changes to allow compilation in CMSSW_9_2_0.  They are designed to have no impact on the code compiling and running in CMSSW_8_0_28.

Biggest changes are:

* auto_ptr changed to unique_ptr everywhere
* std::move added inside put command everywhere
* selectors that are used in templates built off of SingleElementCollectionSelectorPlusEvent now need an operator() method that takes a Ref instead of a const object.  This required some changes to how things were templated, especially the CutBasedGenericObjectSelector 
 and CutBasedElectronObjectSelector
* Special DataFormats xml file to reflect changes in pat objects, which is copied into place by setup_9_2_0.sh
* Special setup_9_2_0.sh file which accesses updated branches

Many thanks to @lgray for helping with an elegant fix to the template issue, which also happens to have been created by him: https://github.com/cms-sw/cmssw/blame/09c3fce6626f70fd04223e7dacebf0b485f73f54/CommonTools/UtilAlgos/interface/SingleElementCollectionSelectorPlusEvent.h#L35

Note that the code will still require changes to run.  The one I'm sure of is removing special EGM id from customization, since those tags haven't been ported yet to CMSSW 9 as far as I know.

### DIFF
--- a/DataFormats/src/classes_def_920.xml
+++ b/DataFormats/src/classes_def_920.xml
@@ -1,0 +1,349 @@
+<lcgdict>
+<class name="flashgg::WeightedObject" ClassVersion="10">
+  <version ClassVersion="10" checksum="1340095011"/>
+</class>
+<class name="flashgg::PDFWeightObject" ClassVersion="14">
+  <version ClassVersion="14" checksum="2888861521"/>
+  <version ClassVersion="13" checksum="3868816395"/>
+  <version ClassVersion="12" checksum="3929296232"/>
+  <version ClassVersion="11" checksum="103726048"/>
+  <version ClassVersion="10" checksum="92077144"/>
+</class>
+<class name="edm::Ptr<flashgg::PDFWeightObject>"/>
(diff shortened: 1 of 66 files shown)


lgray:
@sethzenz If all or most of your code is using edm::Views internally then you don't need to know about the collection type since edm::Views are by required to be transient data products (never stored in event). They merely wrap other things.

In particular, you can make:
```c++ 
template< typename T, template <typename> REF > 
bool operator()( const REF<T>& ref ) const {
   return this->operator(*ref);
}
```
instead and then none of your selectors need to know about the collections anyway.

Also, passing refs by value, not cool!

sethzenz:
@lgray The code didn't compile when I passed them by reference.  I'll investigate if time permits.  Thanks!

sethzenz:
I have tested that none of these changes break the usual tests in 8_0_28.  And the code at least compiles with the 9_2_0 recipe.  So this PR achieves its goals and I'll merge it.

PR #955, opened 2017-11-14
Tag and probe integration in standard flashgg sequence

musella:
Fixed merge issues. Checked compilation in 9_X

Replaces #954


gourangakole:
Hello Pasquale;
I have a small query,
is this PR also compiled in CMSSW_8_0_28?

musella:
@gourangakole The code was developed in 80x, so it compiles there too.

musella:
@sethzenz  it is good to go on my side.

PR #960, opened 2017-12-13
add deep csv in microAOD

malcles:


sethzenz:
Tested in 8_0_28 and 9_2_8.  Runs fine, so I merge now.  Physics performance will of course have to be validated later.

PR #963, opened 2017-12-14
tweak to 80x classdef to restore changes

sethzenz:


sethzenz:
I think I lost a change while integrating #955 and #963 together.  This fixes it.

PR #976, opened 2018-01-23
classes needed to be defined in 94X for tag step

sethzenz:


sethzenz:
Tested: fixes complaint in 94X, has no impact on 80X.

PR #979, opened 2018-01-29
zerothvertexfrommicro

malcles:
Hello Seth,

Committing just small change to be able to have 0th vertex from microAOD.

Cheers,

Julie

### DIFF
--- a/MicroAOD/plugins/DiPhotonProducer.cc
+++ b/MicroAOD/plugins/DiPhotonProducer.cc
@@ -41,7 +41,9 @@ namespace flashgg {
         bool useSingleLeg_;
         unsigned int maxJetCollections_;
         EDGetTokenT<View<reco::GenParticle> >      genPartToken_;
+        bool useZerothVertexFromMicro_;
     };
+
 
     DiPhotonProducer::DiPhotonProducer( const ParameterSet &iConfig ) :
         vertexToken_( consumes<View<reco::Vertex> >( iConfig.getParameter<InputTag> ( "VertexTag" ) ) ),
@@ -56,6 +58,8 @@ namespace flashgg {
         const std::string &VertexSelectorName = iConfig.getParameter<std::string>( "VertexSelectorName" );
         vertexSelector_.reset( FlashggVertexSelectorFactory::get()->create( VertexSelectorName, iConfig ) );
         useSingleLeg_ = iConfig.getParameter<bool>( "useSingleLeg" );
+
+        useZerothVertexFromMicro_ = iConfig.getParameter<bool>( "useZerothVertexFromMicro");
         produces<vector<flashgg::DiPhotonCandidate> >();
     }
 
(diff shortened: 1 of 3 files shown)


sethzenz:
I think in the past there was a solution that simply reevaluated the diphoton momentum using the 0th vertex during processing of microAOD, rather than changing the vertex set in the microAOD.  Has this been checked?

If the current solution is used, what would the MicroAOD workflow be?  Would we have to DiPhoton collections with and without this flag?  Can you check how much extra space that would use?

malcles:
Hi Seth,

In fact in my mind this option is just to be used only to reread micro and reproduce diphotons with 0th vertex on the fly,
 this is why it is called “FromMicro”. But maybe this is not a satisfying solution.
Please let me know.

Julie





On Jan 29, 2018, at 13:11, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


I think in the past there was a solution that simply reevaluated the diphoton momentum using the 0th vertex during processing of microAOD, rather than changing the vertex set in the microAOD. Has this been checked?

If the current solution is used, what would the MicroAOD workflow be? Would we have to DiPhoton collections with and without this flag? Can you check how much extra space that would use?

sethzenz:
If it's updating the microAOD then I would expect it to read a DiPhoton collection and produce a new DiPhoton collection with the changed interpretation.  As written your code reads in photons and makes diphotons, and I'm not sure offhand if it will work running on uAOD for various technical reasons.  But maybe this does work already running on uAOD -- have you tested an "update"-type workflow?

malcles:
Hello Seth,

no I did not try to update the microAOD. I only try to reread microAOD and build up new diphotons on the fly
 and this seems to work. I use it in my config workspaceStd.py-like and I produce ntuples out of it without touching the MicroAOD.

Julie









On Jan 30, 2018, at 09:18, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


If it's updating the microAOD then I would expect it to read a DiPhoton collection and produce a new DiPhoton collection with the changed interpretation. As written your code reads in photons and makes diphotons, and I'm not sure offhand if it will work running on uAOD for various technical reasons. But maybe this does work already running on uAOD -- have you tested an "update"-type workflow?

sethzenz:
OK, I think I guessed wrong from the code structure how this would work.  You're doing exactly the steps I had in mind.  I'll accept this PR now.  When you get the chance can you check in an example configuration that uses this flag just so the situation is clearer?

malcles:
Hello Seth,

Yes I will add an example configuration.

Julie



On Jan 30, 2018, at 10:24, Seth Zenz <notifications@github.com<mailto:notifications@github.com>> wrote:


OK, I think I guessed wrong from the code structure how this would work. You're doing exactly the steps I had in mind. I'll accept this PR now. When you get the chance can you check in an example configuration that uses this flag just so the situation is clearer?

PR #981, opened 2018-02-01
adding DeepCSV discs to flashgg:Jet and adding DeepCSV::probc DeepCSV…

dmajumder:
I'm adding a few missing DeepCSV discriminators to the workflow. 
I'm also adding the DeepCSV discs to the flashgg::Jets as members a la the qglikelihood_. Not sure if this is necessary.

### DIFF
--- a/DataFormats/interface/Jet.h
+++ b/DataFormats/interface/Jet.h
@@ -38,6 +38,18 @@ namespace flashgg {
         float puJetIdMVA() const { return simpleMVA_; }
         Jet *clone() const { return ( new Jet( *this ) ); }
         
+        void  setDeepCSVbb(const float deepcsvbb=-99) {deepcsvbb_ = deepcsvbb;}
+        float DeepCSVbb () const {return deepcsvbb_;}
+        
+        void  setDeepCSVb(const float deepcsvb=-99) {deepcsvb_ = deepcsvb;}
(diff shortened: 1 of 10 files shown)


sethzenz:
I suspect it's not necessary because pat::Jet already has the bDiscriminator method.  Can you check?  We don't want to take up additional space in the jet storing the information twice.

dmajumder:
@sethzenz apologies it took long. I checked and removed the userFloats. If everything else is OK, can you please merge this PR?

sethzenz:
@dmajumder Thanks, this looks ok now, but I have to ask for one thing... can you revert your changes to Taggers/src/LeptonSelection.cc ?   I know their needed to compile in 94X, but the current setup is to keep them compatible with 80X -- the 94X setup script updates them, and I suppose you committed those changes (possibly by accident).  I can sort this out myself, but it's actually easier for you to do it by committing the original file in this branch, if that's ok with you.  Let me know if that needs clarification, and thanks again!

dmajumder:
@sethzenz no problem...was unintentional, I've reverted it.

BTW, is the commit 54486fa OK? Else, I can remove it. But I thought it's a good idea to have a small script to read the MicroAOD contents

dmajumder:
@sethzenz anything that needs to be done here?

PR #985, opened 2018-03-08
updated to Fall17 EGMID

gkrintir:
hi @sethzenz 

The update to Fall17 EGMID.

Successfully tested the microAOD production with 1000 events.

It would be good to update the README, for an afterbuild is temporarily needed.

### DIFF
--- a/DataFormats/interface/Electron.h
+++ b/DataFormats/interface/Electron.h
@@ -30,6 +30,18 @@ namespace flashgg {
         bool passMVAMediumId( ) const {return passMVAMediumId_ ;}
         void setPassMVAMediumId( bool val ) { passMVAMediumId_ = val;}
 
+        bool passMVALooseId( ) const {return passMVALooseId_ ;}
+        void setPassMVALooseId( bool val ) { passMVALooseId_ = val;}
+
+        bool passMVATightNoIsoId( ) const {return passMVATightNoIsoId_ ;}
+        void setPassMVATightNoIsoId( bool val ) { passMVATightNoIsoId_ = val;}
+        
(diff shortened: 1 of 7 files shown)


sethzenz:
@gkrintir Can you please remove your changes to LeptonSelection.cc from this PR?  These changes are made automatically in the 94X setup, and making them permanent in the code will break 80X compatibility.

gkrintir:
hi @sethzenz i spotted that but i thought there is an inverse substitution. 
Shouldn't we define one in setup80X ? Otherwise this should be an issue in every PR, no?

sethzenz:
@gkrintir Only if people making the PR's use "git commit -a" to include every file they changed in the PR. :)   Nevertheless, your point is taken that the inverse substitution in 80X is probably easier.  Let's keep things as they are and I'll do the cleanup of the recipes separately.

In the meantime I tested this and confirmed it runs (although not that any particular output is present).  So I'll accept the PR.

gkrintir:
Ok right with "git commit -a" indeed ;D

Ok nice

michelif:
Hi @gkrintir @sethzenz ,
should also this string be update? this is the string wich configures the branch called EGMPhotonMVA in microAOD, which stores the value of the Egamma MVA ID.
(I know because HHbbgg analysis used that branch).
thanks

gkrintir:
hi @michelif 

right, this is done in the customize part:

https://github.com/cms-analysis/flashgg/blob/master/MicroAOD/python/MicroAODCustomize.py#L384

the file you pointed is simply an initialization ;D

michelif:
ah, ok right.
thanks!

PR #989, opened 2018-03-15
Move diphotons earlier in sequence, stop processing unles pt cut passed

sethzenz:
Improves the previous filtering scheme for data by:
* Requiring diphoton pt > 25,15 GeV for events to be stored (previously, no cut)
* Stopping sequence processing immediately if no such diphoton exists

The former saves space, 20% on a random data file from SingleElectron 2017A.
The latter should save some CPU time (jet, electron, and muon sequences).


sethzenz:
To be checked: whether the code is run in the right sequence for the DoubleMuon case.

PR #991, opened 2018-03-16
Muon sf2017

hbakhshi:
Muon SF 2017 is implemented


sethzenz:
@hbakhshi Can you confirm there's nothing for muon iso or ID that has to be changed at MicroAOD level?

PR #968, opened 2017-12-20
Added extra info for b-jet regression

musella:
- energy in rings for jets
- added gen neutrinos
- leptonPtRel

After addition the first few branches in the event have the following sizes:
File myMicroAODOutputFile.root Events 100
Branch Name | Average Uncompressed Size (Bytes/Event) | Average Compressed Size (Bytes/Event) 
patTriggerObjectStandAlones_selectedPatTrigger__PAT. 76790.7 8690.47
flashggPhotons_flashggRandomizedPhotons__FLASHggMicroAOD. 11072.8 4636.34
flashggElectrons_flashggSelectedElectrons__FLASHggMicroAOD. 4985.69 3217.8
flashggMuons_flashggSelectedMuons__FLASHggMicroAOD. 3227.53 2224.52
flashggJetss_flashggFinalJets__FLASHggMicroAOD. 13487.1 2162.4 ***** Jets are here *******
recoVertexs_offlineSlimmedPrimaryVertices__PAT. 1584.68 896.62

PR #996, opened 2018-03-29
Additional variables for b-jet regression

michelif:
Hi @sethzenz ,
this is an update of pull request #968 by @musella that you already merged before christmas for b-jet energy regression for HH->bbgg. During the review of the  a couple of extra variables were added. I updated the code here with them. Completely neglible changes on size and performance. It would be great if it's merged, so that we can test if everything is ok with the coming production. 
thanks

### DIFF
--- a/MicroAOD/plugins/JetProducer.cc
+++ b/MicroAOD/plugins/JetProducer.cc
@@ -262,7 +262,7 @@ namespace flashgg {
                 std::vector<float> muEnergies(ncone_boundaries+1,0.);
                 int numDaug03 = 0;
 
-                float softLepPdgId=0;
+                int softLepPdgId=0;
 
 
                 for ( unsigned k = 0; k < fjet.numberOfSourceCandidatePtrs(); ++k ) {
@@ -328,7 +328,7 @@ namespace flashgg {
                     fjet.addUserFloat("softLepDr", softLepDr);
                     fjet.addUserFloat("softLepPtRel", softLepPtRel); 
                     fjet.addUserFloat("softLepPtRelInv", softLepPtRelInv); 
-                    fjet.addUserFloat("softLepPdgId", softLepPdgId); 
+                    fjet.addUserInt("softLepPdgId", softLepPdgId); 
                     fjet.addUserInt("numDaug03", numDaug03);
 
                     //                    for(size_t icone = 0; icone < ncone_boundaries+1; ++icone) {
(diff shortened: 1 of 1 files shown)


sethzenz:
@michelif In principle the production is already started as a test (3_0_0), but I can integrate this and use it when submitting further jobs (3_0_1).   Then eventually if a rerun isn't required for another reason, we can go back and replace the 3_0_0 jobs with new 3_0_1's.  I'll work on that now.

In the meantime are there any other changes forseen for HH?  In particular, is there any special configuration to be done for signal samples?  At the moment they are labelled as background samples, hence no PDF information is propagated -- is this OK?   

Finally, can you let me know which MINIAODSIM samples would be high priority for testing this code?  Some HH signals I suppose, but please send the exact sample name(s).

michelif:
Hi Seth, 
thanks. For the final analysis we need this on basically all samples (HH, diphoton, g+Jets, all single higgs, data, t(t)g(g)). It is not super-urgent anyway (we'll have a workshop next week, after that we will have a clearer time scale). For the moment, your idea to have some samples to see that everything and then go back to another reproduction of the samples is ok. 
The easiest sample for you to produce, useful for us, would be ZH since I have two photons and two jets with a mass peak. 
Of course the HH samples would be the best. The exact names looking on DAS are these ones [1], node_SM would be the one highest priority. I remember a mail exchange between you @andreypz and @dmajumder about running on these samples but I don't remember if the problem there was solved. I also let them comment about the PDF (naively I think we should have the PDF info).

[1] https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2F*GluGluToHHTo2B2G*node*%2FRunII*Fall17*%2F*MINIAOD*

thanks

sethzenz:
OK.  I accept this pull request.  I'll launch tomorrow morning as is and include some HH samples.  I'll also include further HGG samples (since I only ran 125 so far).

michelif:
With "all single higgs" I meant all process for M=125 (ggH, VBF, VH, ttH), different masses are not useful for us. thanks again.

sethzenz:
Can you live with 126 for a while for testing purposes?  (I'm just try to avoid immediately rerunning exactly the same samples I already ran.)

michelif:
yes, sure the phase space will not be so different

dmajumder:
Hi, I'm not sure what PDF is used, apart from what's coded here:
https://github.com/cms-analysis/flashgg/blob/master/MicroAOD/python/flashggPDFWeightObject_cfi.py#L9

sethzenz:
@michelif Please see #998 for a catalog containing this change.  Note that only some samples at 126 were available, but I also ran the HH sample.  Let me know if this is sufficient to test or if more is needed.

@dmajumder Some debugging is required to check it's the same pdf name in 2017 and make sure it runs as signal and is stored smoothly.  See also https://twiki.cern.ch/twiki/bin/viewauth/CMS/PDFatCMS   It would be helpful if someone ccan have a look.

PR #1009, opened 2018-04-27
Updated correction files for Legacy2016 and Rereco2017

kmondal:
- Added shower shape correction files for Legacy2016 and Rereco2017

- Added stochastic photon isolation correction for Legacy2016

- flashggUpdatedIdMVADiPhotons_cfi.py is updated with new correction files


sethzenz:
@kmondal The standard tests check out.  I see changes when the photon ID updater is run, which is expected.

Before I accept, can you look into the issue of full5x5_r9 being corrected below 0?  See the flashgg mailing list for discussion.

sethzenz:
Hi @kmondal - is the update here ready?

kmondal:
Hi @sethzenz  as it turns out that pileup reweight was not correct previously for 2017, now I am doing it 
with the updated pileup reweight. I am running a quick final validation now. If everything goes ok, then I will update by tonight. 
Note: only 2017 correction is affected by pileup. Legacy 2016 is okay.

sethzenz:
Tested together with #1017 and changes are only in the places that are expected. Will accept now

Issue #1011, opened 2018-05-02
sigmaEOverE smearing returns inf


sethzenz:
Since #1006 the sigmaEOverE smearing is somehow looking up "inf" for the per-photonresolution, which is then propagated through the central MC resolution and gives infinite-energy photon, which obviously produces surprising results downstream.

I am still checking for how it's happening in detail but here are instructions to reproduce with the current flashgg master:

1. Change the debug to True on this line:

https://github.com/cms-analysis/flashgg/blob/754c8fccb772f025bbc50e6dfa87e05f8d9f79f1/Systematics/python/flashggDiPhotonSystematics_cfi.py#L634

And also on this line:

https://github.com/cms-analysis/flashgg/blob/754c8fccb772f025bbc50e6dfa87e05f8d9f79f1/Systematics/python/flashggDiPhotonSystematics_cfi.py#L709

[and possibly for the 3 other R9 EB/EE combinations]

2. Run the following:

cmsRun Systematics/test/workspaceStd.py doHTXS=1

You will then see error messages like I have...

```
  SigmaEOverESmearing: Photon has pt= 125.474 eta=-0.911612 full5x5_r9=0.949476 and we apply a sigmaEoverE smearing of inf

     sigE/E VALUE BEFORE: 0.00502538

     sigE/E VALUE AFTER: inf

START OF DiPhotonFromPhoton::applyCorrection M PT E1 E2 ETA1 ETA2 160.308 104.375 182.969 35.7221 -0.924076 0.998463

  MCSmearHighR9EBCentral: Photon has energy= 182.969 eta=-0.924076 and we apply a smearing with sigma inf% to get new energy=-inf
```

sethzenz:
This code will catch the issue, and indeed calls an exception on the very first photon:

```
--- a/Systematics/plugins/PhotonSigEoverESmearingEGMTool.cc
+++ b/Systematics/plugins/PhotonSigEoverESmearingEGMTool.cc
@@ -66,6 +66,10 @@ namespace flashgg {
             // computed by getSmearingSigma(...)
             auto sigma = scaler_.smearingSigma(run_number_, y.et(), y.superCluster()->eta(), y.full5x5_r9(), gain, 0, 0);// never apply systematic shift
 
+            if ( sigma < 0. || sigma > 1. ) {
+                throw cms::Exception("SmearingLogic") << " sigmaEOverE is going to be smeared by " << sigma << " which sounds implausible (allowed: 0-1)";
+            }
+
[...]

PR #1012, opened 2018-05-08
fix EGM smearing for missing propagation of photon Et

shervin86:


sethzenz:
Confirmed in a new area that this fixes #1011.  I'll create another PR shortly with a few other tweaks to catch issues like this in the future and make sure the master recipe is clean.

PR #1015, opened 2018-05-15
MiniIsolation has removed

hbakhshi:
Muon MiniIsolation has already included in the official MiniAOD production. Here is the documentation. So It has been removed from the MicroAOD Muon producer.
Have a look at the documentation of Muon in 2017 here 
[SWGuideMuonIdRun2](https://twiki.cern.ch/twiki/bin/viewauth/CMS/SWGuideMuonIdRun2#Muon_selectors_Since_9_4_X)


hbakhshi:
Indeed the MiniIsolation value is not stored in the Muon object. It would
be more clear and safer if the code is parametrized to get the working
point instead of the isolation cut. it is only used by ttHHadronic for some
non standard selection :
https://github.com/cms-analysis/flashgg/blob/fe06c355384cceaae74a0190ca3e82791b29d31f/Taggers/plugins/TTHHadronicTagProducer.cc#L286
if you agree we can ask them to adapt the code and their configurations.
otherwise i can work to use the standard modules of cmssw to read the
miniisolation value and import it into the muon object.

simonepigazzini:
If still relevant please open a new PR on top of the dev_legacy_runII branch.

PR #1017, opened 2018-05-25
update jec, btag sf v2, add 0th vertex example

malcles:


sethzenz:
Tested together with #1009  and changes are only in the places that are expected. Will accept now

PR #1023, opened 2018-05-30
 update ttH hadronic category with new ttHMVA only one category for n…

malcles:
update ttH hadronic category with new ttHMVA including photon variables. only one optimized category for now in the config. possibility to use two categories already included in the code but final 2 optimized categories config to be added tomorrow...


sethzenz:
Superseded by #1025

PR #1031, opened 2018-06-01
fix preselecton cut for tth-hadronic tags

malcles:


sethzenz:
@malcles Looks good.  I'll accept along with any other missing issues and rerun later today

PR #1034, opened 2018-06-01
updated scale and smearing file, modified setup for CMSSW_9_4_6

shervin86:
I have modified the setup instructions, merged and rebased few merge-topics w.r.t. CMSSW_9_4_6 and put the commits in a new branch.


sethzenz:
Previously the setup_9_4_X.sh was organized in blocks, with (in general) 3 components:

1. An explanation of what was about to happen.
1. an addpkg command to put in place the packages that were to be changed by a merge command
1. The merge command itself.

This PR mixes up some blocks and partially removes others, which is inefficient because the packages were added but without any changes.  I am going to start out by removing these entirely to check if we no longer need them:

<verbatim>
-# Straightforward update for 8_0_28	
-echo "Tweaking ConfigToolBase.py to avoid assuming soft link path..."	
-git cms-addpkg FWCore/GuiBrowsers #temp-by hand	
-git cms-merge-topic -u sethzenz:for-flashgg-toolbase-9_4_0	
</verbatim>

<verbatim>
-echo "Setting up weight stuff..."	
-git cms-addpkg CommonTools/UtilAlgos	
-git cms-merge-topic -u sethzenz:for-flashgg-weightscount-9_4_0	
</verbatim>

If it works with those blocks out then they'll stay out, otherwise I'll put them back.  Either way I'll update the setup file to restore the structure.

sethzenz:
Update: the missing branches were included in the new shervin86/for-flashgg-QGL-vertexIndex-9_4_6

sethzenz:
Small changes everywhere but looks ok.  We should do DY validation ASAP but we use this for now to move forward

sethzenz:
One other comment, the external data directories don't get picked up on lxbatch.  I've done a quick workaround for now, we should investigate later

PR #1038, opened 2018-06-20
TH samples added in RunIIFall17-3_1_0 campaign

hbakhshi:
THQ/THW samples with the full set of weight information are added in the RunIIFall17-3_1_0 campaign


sethzenz:
@hbakhshi Thanks for this PR.  It's structured nicely and I think it can be merged easily enough.   To make sure I understand:  What does specifying "WeightName=rwgt_12" do?   Does this cause the SM weight to be used when dumping?

hbakhshi:
@sethzenz Yes, to dump with the Standard Model kinematics weight, one has to set the WeightName parameter to rwgt_12

PR #1046, opened 2018-06-28
Tth lepton bdt

abeschi:
Inclusion of tthLeptonic BDT


sethzenz:
I will run with this PR, but before I accept it needs cleaning.  A lot of files were added that should not have been, e.g. the db files that are moved by the setup script.

sethzenz:
Was the removal of the dilepton tag intended?

PR #1049, opened 2018-07-03
fix tth hadronic defaults

malcles:


malcles:
This fixes the unexpected increased background due to default values that were not set consistently  wrt the ones used in the training and in the optimisation for variables related to 3rd and 4th jets that are not filled for events with 2 or 3 jets.

sethzenz:
Included in #1050 so we can omit this one

PR #1051, opened 2018-07-08
ttH Leptonic BDT

abeschi:
Included latest version of ttH leptonic BDT and new working point.
Updated TTHHadronic tagger with new lepton selections.


sethzenz:
Rejecting because this was replaced by #1053

PR #1050, opened 2018-07-06
Tth hadronic 30vars

malcles:
Hi all, 
these PR includes the latest version of the ttHHadronicMVA (30 variables instead of 35, sensitivity improved by about 8%), with the lastest 3D cuts defined with the optimal cuts found using bkg MC.
It also includes the latest SF for btagging (V3) from the btagging POG. The code has been tested both on Data and MC and is synch with our expectations.

PR #1054, opened 2018-07-09
Tth hadronic 30vars fix

sethzenz:
Should duplicate #1050 after merging with other changes


sethzenz:
I think this replicates #1050 so I proceed but the TTHHadronicProducer in particular should be checked carefully

PR #1059, opened 2018-07-18
simplified code to abbreviate job name to at most 97 characters

andreh7:
removed repeated if statements by a loop over a list


sethzenz:
Thanks!

PR #1072, opened 2018-09-17
overlap remove for ttH leptonic analysis

mhl0116:
Remove overlap between ttGG/ttG/tt sample as discussed in this report:
https://indico.cern.ch/event/754291/contributions/3133922/attachments/1713144/2762621/ttH_gg_overlapRemove_20180910.pdf

to remove overlap events, just require: "lead_Prompt != 0 && sublead_Prompt !=0 "

this code is a developing version and contains several unused members in TTHLeptonicTag.h, can be cleaned up for the future. However, the extra members won't interfere with any of the functionalities of this class


malcles:
Dear Samuel, thanks for this. It is very nice to have the possibility to remove this overlap in flashgg. However, I think it is not mergeable like this in flashgg because it modifies the tags definitions (in flashggTags and also directly in the leptonic tagger). Maybe this was meant only for people to be able to test it. However it would be very nice to have it included permanently in flashgg. So let's discuss this via email. Cheers, Julie

simonepigazzini:
Superseeded by PR #1083 and #1095

PR #1074, opened 2018-09-20
Changes in TnP. 

ArnabPurohit:
(PhotonFromDiPhotonProducer.cc)Added different selections for Leading and subleading. (treeContentPhotons_cfi.py)More probe variables are added.


simonepigazzini:
@ArnabPurohit , @shervin86 , @malcles . Following the last Monday presentation I understood that the TnP study is not yet settled. Should we wait for a more stable TnP or would you like to have this PR integrated?

simone

PR #1075, opened 2018-09-25
Updated ttH leptonic BDT with high stat diphoton sample & updated pre…

abeschi:
New training of the ttHLeptonic BDT with high stat dips sample
Updated BDT thresholds for the ttHLeptonic categories
Updated photon pT thresholds to 35/25 in the preselection module


simonepigazzini:
@beschi, @shervin86 @malcles , this PR changes the pt threshold applied in the diphoton preselection, has this been approved?

abeschi:
The preselection for 2017 should be 35/25 GeV but the value was never updated in the flashgg master.

PR #1079, opened 2018-10-08
fix to have process.dumpPython() working in CMSSW_10_2_1

andreh7:
As discovered by @ArnabPurohit , `process.dumpPython()` fails with CMSSW_10_2_1.

It turns out that this has been found elsewhere in CMSSW (see https://github.com/cms-sw/cmssw/issues/23774 ) but as far as we understood will the fix (https://github.com/cms-sw/cmssw/pull/24168) will only be available in `CMSSW_10_3_X` (according to https://github.com/cms-sw/cmssw/pull/24168#issuecomment-409904682 ). 

We therefore propose a workaround which can be removed once flashgg moves to `CMSSW_10_3_X`.

### DIFF
--- a/MetaData/python/JobConfig.py
+++ b/MetaData/python/JobConfig.py
@@ -350,6 +350,12 @@ class JobConfig(object):
                     target = target.__sub__(lumisToSkip)                    
                 process.source.lumisToProcess = target.getVLuminosityBlockRange()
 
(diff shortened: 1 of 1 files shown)


simonepigazzini:
@andreh7 

Please make the pull request on top of the CMSSW_10_2_X branch and not the master, since the master is at the moment incompatible with 10_2_X.

PR #1081, opened 2018-10-16
HH->bbgg analysis

chernyavskaya:
Working HH->bbgg analysis for 2016 and 2017 eras.


shervin86:
do we really need to keep track of all these commits? It would be better if you can rebase and squash them.

simonepigazzini:
@shervin86 don't worry, we can squash when merging the PR now. I will squash everything into a single commit.

simonepigazzini:
L 994 in MetaData/python/sample_utils.py breaks all sample management tools. I made a quick fix with Nadya (removed copyProxy option) but I suspect that changes to other file were lost before making the PR

MetaData/python/samples_utils.py
```diff
@@ -984,7 +990,8 @@ def __call__(self):
         self.mn = SamplesManager("$CMSSW_BASE/src/%s/MetaData/data/%s/datasets*.json" % (options.metaDataSrc,options.campaign),
                                  dbs_instance=options.dbs_instance,
                                  force=options.doForce,
-                                 queue=options.queue,maxThreads=options.max_threads,doContinue=options.doContinue)
+                                 queue=options.queue,maxThreads=options.max_threads,doContinue=options.doContinue,
+                                 copyProxy=options.copy_proxy)
```
simonepigazzini: @musella @michelif
This breaks all sample management tools. I made a quick fix with Nadya (removed copyProxy option) but I suspect that changes to other file were lost before making the PR

musella: The option should be defined. It was lsot ins some manupulation
```
make_option("--no-copy-proxy",dest="copy_proxy",action="store_false",
                            default=True,help="Do not try to copy the grid proxy to the worker nodes."
),
```

musella:
@simonepigazzini 
I think that the changes should be squashed to 3-4 commits
- b regression
- HH analysis
- fixes to job_utils
- fixes to sample_utils

I guess that we can do the homework and make another PR.

simonepigazzini:
merged in: https://github.com/simonepigazzini/flashgg/commit/8e88c73ac652c4c82a1313c4fdb17729c0b6be50

(to be merged later in this repo)

PR #1107, opened 2019-04-16
EGamma update

michelif:
This PR:
- Sets the correct ID for electrons for 2016/2017/2018
- Sets the correct Scale factors (flashgg code should be improved, not flexible, but keep that for the moment)
- Keeps the possibility to run the old IDs but not run by default
- Some cleaning of double H


MicroAOD/python/MicroAODCustomize.py
```diff
@@ -178,15 +178,15 @@ def customize(self,process):
         if self.runDec2016Regression:
             self.customizeDec2016Regression(process)
         if self.runEGMEleID:
-            #if 'eleIdVersion' in self.metaConditions.keys():
-            getattr(self, 'customize'+self.metaConditions['eleIdVersion'])(process)            
-            # else:
-            #     raise Exception, "runEGMEleID set to True but no eleIdVersion specified in conditions json file"
+            if 'eleIdVersion' in self.metaConditions.keys():
```
simonepigazzini, line 181: Is this for back-compatibility in case one wants to run the old V1 and 2016 IDs?

Systematics/test/workspaceStd.py
```diff
@@ -608,6 +608,10 @@
         print "doing icoll "+str(icoll)
 
         producer = flashggbRegressionProducer.clone(JetTag = coll)
+        producer.bRegressionWeightfile = customize.metaConditions['bRegression']['weightFile']
```
simonepigazzini, line 611: I would prefer to separate b-regression from EGM in two PRs, but ok...

MicroAOD/python/MicroAODCustomize.py
```diff
@@ -326,6 +326,37 @@ def customizeDec2016Regression(self,process):
         process.photonMVAValueMapProducer.srcMiniAOD = cms.InputTag("slimmedPhotons")
         process.photonIDValueMapProducer.srcMiniAOD = cms.InputTag("slimmedPhotons")
 
+    # For the whole RunII use the V2 ID, as explained here  https://indico.cern.ch/event/777545/contributions/3234587/attachments/1766655/2868714/micheli_EG_cmsweek_20181206_4.pdf
```
simonepigazzini, line 329: How stable is this? Should we foresee to have the ID weights and effective areas file configurable from MetaConditions?

simonepigazzini (commented):
Mostly ok. Maybe a separate PR for the changes to the b-regression would be better, keep in mind for the next time

MicroAOD/python/MicroAODCustomize.py
michelif: yes, if one adds a line in the metaconditions can customize this. i thought that someone might want to reproduce the old result.

MicroAOD/python/MicroAODCustomize.py
michelif: if we use this id, it will never change. i copied the structure of the old customize methods. If a new version of the ID will come for ultra-legacy it will have to be updated.

PR #1115, opened 2019-05-07
Recovered past features

abeschi:
Recovering of a few features were lost in the past PR but present in the 2017 ttH analysis:
 - Met correction are now applied
 - BTag scale factors are applied in the correct way
 - Events with 2 leptons can be included in the purest category with a bool flag in flashggTags


simonepigazzini:
@malcles does this PR covers everything you told me was missing from the ttH PAS?

PR #1117, opened 2019-05-16
HHbbgg code compatible with CMSSW 10

nadya-chernyavskaya:
- All recent developments of HHbbgg code
 - Adding additional method to globalVariableComputer to be able to properly pick up and update global variables needed for MVA inputs
 - parallel py modified for psi:  use global protocol instead of infn one to read files 
 - workspaceStd cleaned up from DoubleHbbgg code, reduced to 2 if statements only


DataFormats/src/DoubleHttHTagger.cc
```diff
@@ -0,0 +1,165 @@
...
+                dRLeadJetLepton < dRJetLeptonThreshold || dRSubLeadJetLepton < dRJetLeptonThreshold) continue; 
+
+        goodMuons.push_back( muon );
+    }
+    return goodMuons;
+
+}
+
+
+std::vector<float> DoubleHttHTagger::getVariableValues() const{
```
simonepigazzini: Where is this method used? Is it necessary to return a copy of the object? Could a const reference work as well?

Systematics/test/workspaceStd.py
```diff
@@ -41,18 +41,30 @@
...
                            VarParsing.VarParsing.varType.bool,
                            'doDoubleHTag'
                            )
+customize.options.register('doDoubleHttHKiller',
+                           False,
+                           VarParsing.VarParsing.multiplicity.singleton,
+                           VarParsing.VarParsing.varType.bool,
+                           'doDoubleHttHKiller'
+                           )
+customize.options.register('ttHKillerInputVariables',
```
simonepigazzini: I found confusing to call a singleton, bool option "InputVariables". What is this used for?

Taggers/python/flashggDoubleHReweight_cfi.py
```diff
@@ -1,8 +1,22 @@
 import FWCore.ParameterSet.Config as cms
 
+#default values first
+weightsFile = "flashgg/Taggers/data/HHTagger/HHreweight_2016nodes_04022018.root"
```
simonepigazzini: Is this a joke? Why don't you just use the metaConditions?

simonepigazzini: same later on ...

nadya-chernyavskaya: My mistake, I forgot to put default values for reweighting to the MetaConditions. Fixed now. I also set all default values in the config files to empty. In this case if MetaConditions are not properly filled, it will crash.

Systematics/test/workspaceStd.py
nadya-chernyavskaya: It is used to save input variables needed for tthKiller retraining . I renamed it to ttHKillerSaveInputVariables

DataFormats/src/DoubleHttHTagger.cc
nadya-chernyavskaya: I could not find where this method is used, I think this is outdated. I removed it (in the meantime will verify with Thong)

simonepigazzini:
Thanks, do you expect further developments in the next week?

nadya-chernyavskaya:
No, we are not planning any developments in the next week.

simonepigazzini:
Thank you for the changes, merging now

PR #1119, opened 2019-05-23
Update handling of jets in microAOD

edjtscott:
This PR simply updates the jet treatment in microAOD to be the same as it already was in 2017. Performed tests to check that the zeroth vertex jets agree with the miniAOD jets, and that the jets from the other vertices are sensible. These tests were done on 2016 & 2018 miniAOD for data, Drell-Yan and signal. 

The second commit has two minor adjustments that were needed to get these local tests to work - I don't know if they will be required or wanted in the main repo. 

At some point we expect further updates for the QGL and PUJID, but for now the code should be up-to-date.

### DIFF
--- a/MicroAOD/python/flashggJets_cfi.py
+++ b/MicroAOD/python/flashggJets_cfi.py
@@ -92,7 +92,7 @@ def addFlashggPFCHSJets(process,
     process.QGPoolDBESSource = cms.ESSource("PoolDBESSource",
                                             CondDBSetup,
                                             toGet = cms.VPSet(),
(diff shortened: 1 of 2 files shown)


simonepigazzini:
Hi Ed,

thanks a lot. Can you please later add here a link to the validation plots? I will also undo the changes you made to fix the crash, the issue is that the setup script does not take care of copying the QGL db file under flashgg/ and so the fix has to be made in the setup script rather than in flashggJets_cfi.py (unfortunately the QGL implementation is not very smart and this dirty trick is the only one that makes it work both locally and on the GRID).

simone

PR #1121, opened 2019-06-13
ttH Run II MVAs

sam-may:
Implementation of MVA functions for ttH Hadronic and ttH Leptonic channels. As shown in https://indico.cern.ch/event/812024/contributions/3462723/attachments/1861401/3059302/ttH_gg_update_20190610_final.pdf


Taggers/python/flashggTags_cff.py
```diff
@@ -63,16 +63,20 @@
                                        rhoTag = cms.InputTag('fixedGridRhoFastjetAll'),
 #                                       tthMVAweightfile = cms.FileInPath("flashgg/Taggers/data/TMVAClassification_tth_hadronic_2017Data_35vars_v0.weights.xml"),
                                        tthMVAweightfile = cms.FileInPath("flashgg/Taggers/data/TMVAClassification_tth_hadronic_2017Data_30vars_v0.weights.xml"),
+				       topTaggerXMLfile = cms.FileInPath("flashgg/Taggers/data/resTop_xgb_csv_order_deepCTag.xml"),
```
simonepigazzini, line 66: Are these trainings year dependent? If so we should steer the file name from the metaconditions.

Taggers/python/flashggTags_cff.py
```diff
@@ -176,12 +181,16 @@
                                        GenParticleTag=cms.InputTag( "flashggPrunedGenParticles" ),
 				       rhoTag = cms.InputTag('fixedGridRhoFastjetAll'),
                                        MVAweightfile = cms.FileInPath("flashgg/Taggers/data/TMVAClassification_BDT_training_v2.json.weights.xml"),
-                                       leadPhoOverMassThreshold = cms.double(0.33),
-                                       subleadPhoOverMassThreshold = cms.double(0.25),
-                                       MVAThreshold = cms.vdouble(0.6, 0.3),
-                                       PhoMVAThreshold = cms.double(-0.2), 
+				       topTaggerXMLfile = cms.FileInPath("flashgg/Taggers/data/resTop_xgb_csv_order_deepCTag.xml"),
```
simonepigazzini, line 184: same here

setup_flashgg.sh
```diff
@@ -79,6 +79,11 @@ cp XGBoostCMSSW/XGBoostInterface/toolbox/*xml $CMSSW_BASE/config/toolbox/$SCRAM_
 scram setup rabit
 scram setup xgboost
 
+# Grab xml file for top-tagger BDT from ttH MultiLepton analysis (too large to store in Github)
+pushd flashgg/Taggers/data/
+wget "http://uaf-8.t2.ucsd.edu/~sjmay/ttH/resTop_xgb_csv_order_deepCTag.xml"
```
simonepigazzini, line 84: I am considering moving all the weights files to EOS keeping only a sym-link in the repo. Can you easily access EOS from your T2?

MicroAOD/python/flashggDiPhotons_cfi.py
```diff
@@ -12,8 +12,8 @@
                                   GenParticleTag         = cms.InputTag( "flashggPrunedGenParticles" ),
 
                                   ##Parameters for Legacy Vertex Selector                                                
-                                  vertexIdMVAweightfile   = cms.FileInPath(""),
-                                  vertexProbMVAweightfile = cms.FileInPath(""),
+                                  vertexIdMVAweightfile   = cms.FileInPath("flashgg/MicroAOD/data/TMVAClassification_BDTVtxId_SL_2016.xml"),
```
simonepigazzini: You should get the weight file name from meta conditions, where it is already available.

simonepigazzini (changes requested):

Taggers/python/flashggTags_cff.py
sam-may: Currently, no -- there is a single training for all 3 years. I think it is likely that we will keep it as a single training for all 3 years, and add the year information as an input to the BDTs.

Taggers/python/flashggTags_cff.py
sam-may: Same here as well: single training for all 3 years.

setup_flashgg.sh
sam-may: Yes, that would be a better solution than what I've done. I believe I can access /eos easily from the T2 (I'll let you know if we're unable to)

MicroAOD/python/flashggDiPhotons_cfi.py
sam-may: Ok, makes sense. I will change this.

simonepigazzini: thanks, I will also make a script to make the weight file "persistent" by copying them from EOS in case EOS availability is an issue.

sam-may:
I added a commit removing the hard-coded values in `flashggDiPhotons_cfi.py`.

I think for now it makes sense to keep the top-tagger and BDT/DNN trainings in `flashggTags_cff.py`, since we are not currently planning to make these year-dependent. If we do make them year-dependent, I'll move them to the MetaConditions, as you suggest.

Let me know if there's anything else that needs to be changed.

simonepigazzini:
I agree, I will merge this later today/tomorrow.

simone

simonepigazzini (approved):

PR #1130, opened 2019-07-12
Met Filters for Run II

sam-may:
* Application of MET filters now performed through `flashggMetFilters` module.
* All filters, except 1, are applied at the workspace production level.
* List of filters is configurable by year and for data/MC. Currently set to the recommendations provided in [1].
* The one filter that is not applied at workspace production level is done at microAOD production level because it must be rerun on top of miniAOD. I follow the recipe given in [1].

[1] https://twiki.cern.ch/twiki/bin/view/CMS/MissingETOptionalFiltersRun2#Analysis_Recommendations_for_ana

### DIFF
--- a/MicroAOD/test/microAODstd.py
+++ b/MicroAOD/test/microAODstd.py
@@ -17,15 +17,15 @@ process.MessageLogger.cerr.FwkReport.reportEvery = cms.untracked.int32( 100 )
 import os
 ### 2016
 process.GlobalTag = GlobalTag(process.GlobalTag, '', '')
-process.source = cms.Source("PoolSource",
-                             fileNames=cms.untracked.vstring("/store/mc/RunIISummer16MiniAODv3/VBFHToGG_M125_13TeV_amcatnlo_pythia8_v2/MINIAODSIM/PUMoriond17_94X_mcRun2_asymptotic_v3-v1/50000/38128C3C-892D-E911-AC8E-008CFA0087C4.root"))
+#process.source = cms.Source("PoolSource",
+#                             fileNames=cms.untracked.vstring("/store/mc/RunIISummer16MiniAODv3/VBFHToGG_M125_13TeV_amcatnlo_pythia8_v2/MINIAODSIM/PUMoriond17_94X_mcRun2_asymptotic_v3-v1/50000/38128C3C-892D-E911-AC8E-008CFA0087C4.root"))
 #    process.GlobalTag = GlobalTag(process.GlobalTag,'80X_dataRun2_2016LegacyRepro_v4','')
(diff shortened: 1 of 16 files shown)


MicroAOD/test/microAODstd.py
```diff
@@ -147,4 +147,7 @@
 if "DY" in customize.datasetName or "SingleElectron" in customize.datasetName or "DoubleEG" in customize.datasetName or "EGamma" in customize.datasetName:
     customize.customizeHLT(process)
 
+if "Era2017" in customize.options.conditionsJSON or "Era2018" in customize.options.conditionsJSON:
```
simonepigazzini: Nope. Add something like "METFilters" to the meta conditions and then read them from here. Otherwise if something will change from ReReco to UL we will forget to switch this off ...

MicroAOD/test/microAODstd.py
```diff
@@ -89,18 +89,43 @@
...
-
-process.BadChargedCandidateFilter.muons = cms.InputTag("slimmedMuons")
-process.BadChargedCandidateFilter.PFCandidates = cms.InputTag("packedPFCandidates")
-process.BadPFMuonFilter.muons = cms.InputTag("slimmedMuons")
-process.BadPFMuonFilter.PFCandidates = cms.InputTag("packedPFCandidates")
-
-process.flag_globalTightHalo2016Filter = cms.Path(process.globalTightHalo2016Filter)
-process.flag_BadChargedCandidateFilter = cms.Path(process.BadChargedCandidateFilter)
-process.flag_BadPFMuonFilter = cms.Path(process.BadPFMuonFilter)
+# Disable filters below, now applied through flashggMetFilters module in workspaceStd.py
```
simonepigazzini: If this piece of code is no longer needed remove it instead of commenting it please. Git will keep track of it for us :)

MicroAOD/test/microAODstd.py
sam-may: Yes makes sense, I see your point. I added this as a `bool` option in the MetaConditions (`rerun_ecal_calib_filter`).

I also moved the implementation of the filter to `MicroAODCustomize.py`, as this way the filter module is then also included in the printed out sequence.

MicroAOD/test/microAODstd.py
sam-may: Removed these as well as some other unnecessary ones in `workspaceStd.py`

sam-may:
Hi @simonepigazzini @malcles , 

Julie pointed out that it's not optimal to have the `ecalBadCalibFilter`, which has to be rerun on top of miniAOD, be applied at microAOD production level. This is because it will filter events even for analyses that don't use MET.

So instead, I added a member variable to the Met data format which stores the result of the filter. Tags for which Met is important can then decide at workspace production level if they want to cut on the filter or not.

simonepigazzini:
Shall I merge this? @malcles @shervin86

PR #1137, opened 2019-08-05
Modified systematics workflow for ttH tags

sam-may:
Implement modified systematics workflow for ttH tags to prevent excessive memory use by the large MVAs used by the ttH taggers. This is only an issue when running with full systematics (`doSystematics=True`), as an instance of each tagger object is created for each systematic variation by default.

- When running with `tthTagsOnly=True`, evaluates all systematic variations in a single instance of the tagger object, looping over all relevant variations. Reduces memory use from 50GB -> 3GB.
- When running with `tthTagsOnly=False`, disables large MVAs in the ttH tags (unless `doSystematics=False`, in which case full MVAs are still used). The systematics are run in the original way here, as I have not yet determined how to modify the systematics workflow for the ttH tags while maintaining the original workflow for the other tags.
- @mhl0116 and I have verified that this modified workflow produces identical results as the original systematics workflow and that jobs no longer take up excessive memory leading to crashes (tested both locally and in batch with `htcondor` on `lxplus`).
- To-do: how can we implement the memory-intensive MVAs when running all tags with the full systematics sequence? It seems like we would need to modify all taggers to loop over systematic variations inside the tagger. Or maybe there is an easier solution that I'm missing?


simonepigazzini:
Hi Sam,

thank you for this PR. Concerning your last bullet I think there is no easy way, we should rewrite the taggers such that they deal internally with the sys variations.

I'll take a look at the PR and merge it during this week.

simone

simonepigazzini:
@sam-may, can I merge this PR or are you still working on it?

sam-may:
Hi @simonepigazzini , sorry for the late reply, I was away on vacation. @mhl0116 and I found that there was a bug with some systematics being assigned to the wrong tags. Let me first do some more rigorous checks to make sure everything is fixed.

I will confirm with you tomorrow.

sam-may:
Hi @simonepigazzini , @mhl0116 and I finished running some more detailed checks. We compared the m_gg distributions for every systematic in every ttH category, running over ~24k events, with both the old workflow and the new workflow [1]. The systematics match exactly between the two, with the exception of a single event which we believe is due to floating point rounding: TTHHadronicTag_3 has 2673 vs. 2674 events, all other tags match exactly.

So, the PR is ready to be merged now. Thanks.

[1] http://uaf-8.t2.ucsd.edu/~sjmay/2019/20190819_compareShape/

simonepigazzini:
Hi Sam,

thank you, in particular for the detailed checks. Could you please fix the conflict that arose after merging other PRs? Should not take much.

simone

PS: you may want to have a look at this https://github.com/simonepigazzini/php-plots, to improve the readability of your web pages (if you don't have something similar already).

sam-may:
Conflicts are fixed now. And thanks for the link, the plots should be easier to browse now :)

PR #1140, opened 2019-08-22
Load parent and secondary dataset options in JobConfig

simonepigazzini:
- useParentDataset: load miniAOD collection while running on microAODs, add example for STXS 1.1 bin re-evaluation
- secondaryDataset: load collections from user specified dataset (needed by H->4g analysis).

The two options are exclusive, useParentDataset has been tested by Ed and Jonathon while secondaryDataset might need further debug.


simonepigazzini:
Just a test PR. Will be actually merged later.

PR #1148, opened 2019-09-18
Add option for recalculating PDF weights on-the-fly

sam-may:
Adding the `mc2hessianCSV` file and rerunning the `flashggPDFWeightObject` fixes the issue of missing PDF weights in the `Era2017_RR-31Mar2018_v2` and `Era2018_RR-17Sep2018_v2` microAOD campaigns.

### DIFF
--- a/Systematics/python/SystematicsCustomize.py
+++ b/Systematics/python/SystematicsCustomize.py
@@ -301,3 +301,26 @@ def runRivetSequence(process, options):
                                          signalParticlePdgIds = cms.vint32(25), ## for the Higgs analysis
                                      )
     process.p.insert(0, process.mergedGenParticles*process.myGenerator*process.rivetProducerHTXS)
+
+def recalculatePDFWeights(process, options):
+    print "Recalculating PDF weights"
+    process.load("flashgg/MicroAOD/flashggPDFWeightObject_cfi")
+    process.flashggPDFWeightObject = cms.EDProducer('FlashggPDFWeightProducer',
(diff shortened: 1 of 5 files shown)


Systematics/python/SystematicsCustomize.py
```diff
@@ -301,3 +301,26 @@ def runRivetSequence(process, options):
...
+                                                    nPdfEigWeights = cms.uint32(60),
+                                                    mc2hessianCSV = cms.untracked.string(""),
+                                                    LHERunLabel = cms.string("externalLHEProducer"),
+                                                    Debug = cms.bool(False),
+                                                    PDFmap = cms.PSet(#see here https://lhapdf.hepforge.org/pdfsets.html to update the map if needed
+                                                        NNPDF30_lo_as_0130_nf_4 = cms.untracked.uint32(263400),
+                                                        NNPDF31_nnlo_as_0118_nf_4 = cms.untracked.uint32(320900)
+                                                    )
+                                                ) 
+    setattr(process.flashggPDFWeightObject, "mc2hessianCSV", "PhysicsTools/HepMCCandAlgos/data/NNPDF30_lo_as_0130_hessian_60.csv")
```
simonepigazzini: I think it's best to put the path into the MetaConditions such that we also get the correct value in the microAOD the next time we run them

sam-may: Makes sense, I will update this.

sam-may: Updated to accessing the `mc2hessianCSV` file from the MetaConditions in the most recent commit.

PR #1151, opened 2019-09-21
Implementing L1 Pre-firing weights

sam-may:
Implemented L1 pre-firing inefficiencies for ttH tags. 

The pre-firing module will only be run if you select `doL1Prefiring = True` when running `workspaceStd.py`. If that option is selected, the `MetaConditions` file is then checked to see if pre-firing is appropriate for that year (i.e. don't apply if it is 2018), and if so, the weight files are set accordingly.

The effect is on the order of a few % for 2016 and 2017. Below are comparisons of workspaces with and without the L1 pre-firing weights applied. As a sanity check, 2018 is unchanged whether we select `doL1Prefiring = True` or `doL1Prefiring = False`. These were just quick checks on 1000 events in a `ttHJetToGG_M125` file for each year, and we shouldn't quote any efficiency numbers from these.

I copied all the necessary weight files to `/eos/cms/store/group/phys_higgs/cmshgg/flashgg-data/Taggers/data/`.

2016: http://uaf-10.t2.ucsd.edu/~sjmay/ttH/L1Prefiring_comparison_2016/
2017: http://uaf-10.t2.ucsd.edu/~sjmay/ttH/L1Prefiring_comparison_2017/
2018: http://uaf-10.t2.ucsd.edu/~sjmay/ttH/L1Prefiring_comparison_2018/


sam-may:
Hi Simone, I cleaned up the config a little bit with the last commit. The L1Prefire module was being kept in the sequence unnecessary cases, like when running on data, sometimes causing crashes. I made it more "user-proof" and ensured that if the user selects `doL1Prefiring=True` it only actually gets run and has the weights applied if it makes sense to do so (i.e. only for MC, only for 2016/2017).

simonepigazzini:
@sam-may @edjtscott , I am wondering if it wouldn't be better to have the prefiringProducer to produce a new dipho collection with the weight applied instead of relying on each tag producer to apply it itself.
What do you think? 

simone

edjtscott:
Yes I agree that would make more sense. It was set up this way for the use case where one might want to dump the weight but not necessarily apply it to the diphoton object by default. However I don't think we really need that, and am sure we could get that working with a modified dipho collection anyway. 

I am happy to make the necessary changes but would be on timescale of ~2weeks. I guess @sam-may can work with the current config til then?

sam-may:
@simonepigazzini @edjtscott , yes, @mhl0116 and I are fine using the current setup for the results now, and I agree that Simone's suggestion sounds better.

Ed, let me know if there's anything I can help with for this.

edjtscott:
I've modified the prefire producer as planned - @sam-may could you please check that it still behaves correctly for your use cases? If it does then I will perform the rebase before @simonepigazzini checks all is suitable.

sam-may:
Hi @edjtscott , thanks for making these updates. I took a look at the logic for each year and data vs. MC -- everything looks correct to me.

I merged the changes and tried building and running some test cases, but I seem to always get a segmentation fault from L68 of `Taggers/plugins/PrefireDiPhotonProducer.cc`. I'm guessing I've just made a mistake somewhere in merging the new changes, I will take a look at debugging this more tomorrow.

simonepigazzini:
Hello, how things are going here?

simone

sam-may:
Hi Simone, I haven't had a chance to debug this yet (busy with review comments for ttH), I will take a look at this again soon.

At that point I can also add the associated systematic uncertainties as detailed in the TWiki.

Let me know if this sounds good to you, @simonepigazzini and @edjtscott

simonepigazzini:
@sam-may is the issue solved? can I merge this?

sam-may:
Hi @simonepigazzini , no, I still haven't implemented the systematic uncertainties following the prescription in the TWiki. I'll let you know when I finish implementing this.

sam-may:
Hi @simonepigazzini and @edjtscott , I've added systematic uncertainties for L1 prefiring following the official recipe [1]. The up/down variations for the prefiring probability are saved in workspaces following the same pattern as the other systematics and the values look reasonable after some checks.

Let me know if this looks good to you. If so, I will resolve the conflicts.

[1] https://twiki.cern.ch/twiki/bin/viewauth/CMS/L1ECALPrefiringWeightRecipe

simonepigazzini:
Hi, it looks good. The only issue I have is that the prefiring weights are applied only in the ttH taggers. @Ed shall we move this into the tag sorter as well?

Since the PR has conflicts I would rather merge the STXS PR first and then deal with the conflicts here only once.

edjtscott:
> Hi, it looks good. The only issue I have is that the prefiring weights are applied only in the ttH taggers. @ed shall we move this into the tag sorter as well?
> 
> Since the PR has conflicts I would rather merge the STXS PR first and then deal with the conflicts here only once.

Actually I changed the implementation of this to produce a new collection of diphotons with the weights added, and left it configurable whether to apply the weight to the central value. So there should be no need to apply the weights separately in the individual taggers. 

Also, I think it makes sense to have the new collection rather than the TagSorter since the eventual weight depends on the jet collection, which in turn depends on the vertex. So if I have understood correctly we can just remove the lines in the ttH taggers, rebase on top of the latest then all good to merge.

edjtscott:
Hi @sam-may (cc @simonepigazzini) shall I fix this and merge the latest changes, or do you prefer to do it? From my side it would be some time next week

sam-may:
Hi @edjtscott , sorry got distracted from this, I can do this today.

sam-may:
Hi @edjtscott and @simonepigazzini , the changes from the original implementation that was specific to the ttH tags have been reverted now.

edjtscott:
ok great thanks - have you tried merging the latest "master" & resolving the conflicts?

sam-may:
I've now merged in the latest version of `dev_legacy_runII` and resolved the conflicts.

I've also tested that it still runs properly -- it does, but I had to remove a couple parts. The first is just a single line (commit `aa45d1c`) and the second (`83d00e4`) is related to the overlap removal code the @mhl0116 implemented a while ago.

The overlap removal is only relevant for background MC (i.e. only relevant for personal studies, not any official results), so I think it's fine to comment it out for now. It's unclear to me what exactly is causing the crash, but it seems to be a bad pointer to the `genParticles` in the tagger. The error message [1] comes from this line [2] in which we start a loop over the gen particles. This worked fine in CMSSW_10_5_0 (and before merging STXS stuff), but now crashes.

[1] `flashgg::TTHLeptonicTagProducer::GenPhoIndex (this=this@entry=0x7f2ee818c000, genParticles=..., pho=pho@entry=0x7f2edef03000, usedIndex=usedIndex@entry=-1) at /home/users/sjmay/ttH/forPR_topicL1Prefire_v3/CMSSW_10_6_1_patch2/src/flashgg/Taggers/plugins/TTHLeptonicTagProducer.cc:351`
[2] https://github.com/cms-analysis/flashgg/blob/dev_legacy_runII/Taggers/plugins/TTHLeptonicTagProducer.cc#L351

edjtscott:
Hi Sam thanks a lot for sorting that. I’m pretty sure the crash relating to the genParticles is because I accidentally removed the corresponding getByToken line in the STXS PR. If you can check and add it back in that’s perfect, otherwise I will tomorrow. 

Otherwise, unless there are additional comments from @simonepigazzini, I think we’re good to merge this.

sam-may:
You were right Ed, that fixed it :)

simonepigazzini:
Everything looks good, merging now.

Thank you for implementing this and taking care of resolving the conflicts.

simone

PR #1153, opened 2019-09-30
Added check for data in PhoIdInputsCorrector

threiten:


simonepigazzini:
@sam-may @mhl0116 you need this fix for data. My bad, I thought of putting the check for real data in the python but changed my mind eventually without fixing the c++, we were correcting data too :(

simone

PR #1163, opened 2019-10-23
Malcles/fix pixel seed sf

malcles:
Here are the new pixel seed SF. I will now try to add them to the workspaceStd and check it works but numbers at least are available here. 
They are based on Junquan SF from z->mumugamma:
 https://indico.cern.ch/event/850506/contributions/3606914/attachments/1927946/3192290/201910_Zmmg_RhoBugFixed.pdf
Plus ttH pixel seed efficiencies (after PU reweighting) to make them for both passing and non passing events while preserving total number of events (it was checked that eff are very similar also for ggH). These are to be used only for tags using hasPixelSeed info.

### DIFF
--- a/MicroAOD/python/flashggDiPhotons_cfi.py
+++ b/MicroAOD/python/flashggDiPhotons_cfi.py
@@ -12,8 +12,8 @@ flashggDiPhotons = cms.EDProducer('FlashggDiPhotonProducer',
                                   GenParticleTag         = cms.InputTag( "flashggPrunedGenParticles" ),
 
                                   ##Parameters for Legacy Vertex Selector                                                
-                                  vertexIdMVAweightfile   = cms.FileInPath(""),
-                                  vertexProbMVAweightfile = cms.FileInPath(""),
+                                  vertexIdMVAweightfile   = cms.FileInPath("flashgg/MicroAOD/data/TMVAClassification_BDTVtxId_SL_2016.xml"),
+                                  vertexProbMVAweightfile = cms.FileInPath("flashgg/MicroAOD/data/TMVAClassification_BDTVtxProb_SL_2016.xml"),
 
                                   useSingleLeg            = cms.bool(True),
                                   useZerothVertexFromMicro = cms.bool(False),
--- a/MicroAOD/python/flashggDiPhotons_cfi.py
+++ b/MicroAOD/python/flashggDiPhotons_cfi.py
@@ -12,8 +12,8 @@ flashggDiPhotons = cms.EDProducer('FlashggDiPhotonProducer',
                                   GenParticleTag         = cms.InputTag( "flashggPrunedGenParticles" ),
 
                                   ##Parameters for Legacy Vertex Selector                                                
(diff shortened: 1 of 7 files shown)


simonepigazzini:
@mhl0116 @sam-may have you already merged this into your working branch? Shall I close this and wait from a PR on your side with the latest changes? Also it's not clear to me if this additional systematics is loaded correctly if we run the ttH tag together with all the others, have you ever checked the behavior in that scenatio?

Thank you,

simone

malcles:
Hi Simone,

No I have to say that the behaviour with all tags in was not checked, this version works only for the option ttHTagsOnly=true.

Julie



On 13 Nov 2019, at 10:46, simonepigazzini <notifications@github.com<mailto:notifications@github.com>> wrote:


@mhl0116<https://github.com/mhl0116> @sam-may<https://github.com/sam-may> have you already merged this into your working branch? Shall I close this and wait from a PR on your side with the latest changes? Also it's not clear to me if this additional systematics is loaded correctly if we run the ttH tag together with all the others, have you ever checked the behavior in that scenatio?

Thank you,

simone

mhl0116:
Hi @simonepigazzini 

I merged these commits by hand from @malcles  branch, they are in the latest result with ttHTagsOnly=true. Not sure how it will behave if switching this into false as Julie said. For the similar reason, we probably need a separate discussion on how should we push our working branch: we are not 100% sure if the working branch won't cause any unexpected flashgg behavior for others if switching ttHTagsOnly into false.

Best,
Hualin

malcles:
For the pixelSeedWeight, as implemented now it should not prevent other tags to work when tthTagsOnly is switched off
 but this weight will not be included in the ttHtags while running the full tag sequence, this will need to be changed.
Hualin, in any case when you will have more time it would be great to commit all the needed changes
to flashgg so that ttH results can be reproduced.  And we will need to think about how to make ttH tags work
with the full tag sequence in, since there are also more complex modifications to integrate (systematics workflow for example).

Best,
Julie

On 13 Nov 2019, at 13:35, Hualin Mei <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi @simonepigazzini<https://github.com/simonepigazzini>

I merged these commits by hand from @malcles<https://github.com/malcles> branch, they are in the latest result with ttHTagsOnly=true. Not sure how it will behave if switching this into false as Julie said. For the similar reason, we probably need a separate discussion on how should we push our working branch: we are not 100% sure if the working branch won't cause any unexpected flashgg behavior for others if switching ttHTagsOnly into false.

Best,
Hualin

simonepigazzini:
@mhl0116 @sam-may @edjtscott has this PR been merged into one of your other PRs? If so does the standard sequence correctly read the PSV SF for ttH?

thank you,

simone

mhl0116:
This PR has not been merged into other PRs, and it would only load PSV SF for ttH if setting tthTagsOnly to True. We need to modify this to work with latest STXS workflow when ttH runs with other tags.

PR #1167, opened 2019-11-08
HHbbgg analysis with full Run II data

nadya-chernyavskaya:
HHbbgg analysis with full Run II data


simonepigazzini:
Dear @fabio-mon and @chernyavskaya ,

the HH part of the PR looks ok, but I have mejor concerns about the ttH+HH part some of them have been pointed out with inline comments, I want to stress the following:

1) The classes used to compute the on the fly reweighting can, and should, be compiled through scram.
    Having a separate makefile (which is not even a makefile but a script) is a recipe for a disaster.
2) We should discuss the details in person but having a list of microAOD file to process is also something that we want to avoid (let alone the file with a nanoAOD list!). We should use the catalogues and JobConfig for that. With fggRunJobs you can use it and you can also specify any executable or script to be run.
3) There hare hardcoded file names and commets that need a cleanup
4) This is low priority but we should move the gen info that are now in the DoubleHTag into the GenDiphoton object.

simone

simonepigazzini:
Ok discussing with Nadya it emerged that some of the scripts added here are supposed to be added to flashggFinalFits and not to flashgg so we will close this PR and work on a cleaned up one.

simonepigazzini:
Superseded by PR #1178

PR #1168, opened 2019-11-13
Fix extraction of [secondary]Dataset files from DAS in JobConfig

mbandrews:
In extraction of secondaryDataset files in `MetaData/python/JobConfig.py`, `runs_and_lumis[str(s)]` was only extracting the first run in each file `s` from the DAS query,

`{ ... for lumi in das_query(...)['data'][0]['lumi']}`

but `len(das_query(...)['data'])` == number of runs contained in file `s`, leading to missing files when running with a secondary dataset. This has been fixed to extract all runs in the file,

`{ ... for data in das_query(...)['data']}`

and similarly for the extraction of primary dataset files in `f_runs_and_lumis`.

### DIFF
--- a/MetaData/python/JobConfig.py
+++ b/MetaData/python/JobConfig.py
@@ -381,18 +381,18 @@ class JobConfig(object):
                 process.source.lumisToProcess = target.getVLuminosityBlockRange()
 
                 print process.source.lumisToProcess
-            
+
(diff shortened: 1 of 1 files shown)


simonepigazzini:
Thank you for the fix Michael. It's merged.

simone

PR #1169, opened 2019-11-13
Add stage 1 STXS workflow to dev branch

edjtscott:
This PR contains the new tag structure developed by Seth and me for the stage 1 STXS measurements. All that needs to be done is to set doStageOne=True in workspaceStd; has been checked that nothing breaks without this option. So far tested with 2017 data & MC only. 

To be addressed before or after merging:
- Some systematics still missing from output workspaces, to be understood
- Understand how best to incorporate with other workflows developed in the meantime (such as ttH)


MicroAOD/python/flashggJets_cfi.py
```diff
@@ -251,5 +251,11 @@ def addFlashggPuppiJets(process,
                                   inputTagJets= PuppiJetCollectionVInputTag
 )
 
-  
+flashggUnpackedJets = cms.EDProducer("FlashggVectorVectorJetUnpacker",
```
simonepigazzini, line 254: What does this change imply?

Systematics/test/data_2017.json
```diff
@@ -0,0 +1,13 @@
+{
```
simonepigazzini: Maybe move all the json files and script under test/runHTXS or similar?

Systematics/test/workspaceStd.py
```diff
@@ -495,9 +501,11 @@
               currentVariables = systematicVariablesHTXS
           else:    
               currentVariables = systematicVariables
-      if tagName == "NoTag":
+      if tagName.upper().count("NOTAG"):
```
simonepigazzini, line 529: Should be ok but @chernyavskaya might want to double check.

Taggers/interface/CollectionDumper.h
```diff
@@ -367,14 +377,13 @@ namespace flashgg {
                 }
             }
         }
-        if (splitPdfByStage0Cat_ && dumpPdfWeights_) {
+        if (reweighGGHforNNLOPS_) {
```
simonepigazzini, line 380: @threiten this might come in handy for differentials as well.

Taggers/python/flashggDiPhotonMVA_cfi.py
```diff
@@ -1,36 +1,13 @@
...
-                                    
-                                    # xml 2015                                    
-                                    #            VertexProbParamsConv=cms.vdouble(-0.049,-0.241,-0.505,-0.270),
-#                                   VertexProbParamsNoConv=cms.vdouble(-0.344,-0.091,-0.234,-0.186),
-                                    # xml 2016 fix low nvtx
                                     VertexProbParamsConv=cms.vdouble(-0.045,-0.148,-0.328,-0.184),
                                     VertexProbParamsNoConv=cms.vdouble(-0.366,-0.126,-0.119,-0.091),
                                     doSigmaMdecorr = cms.bool(True),
-#                                    sigmaMdecorrFile = cms.FileInPath("flashgg/Taggers/data/diphoMVA_sigmaMoMdecorr_split_80x_new.root"),
                                     sigmaMdecorrFile = cms.FileInPath("flashgg/Taggers/data/diphoMVA_sigmaMoMdecorr_split_Mgg40_180.root"),
```
simonepigazzini, line 11: @threiten @vtavolar we should probably make this configurable from MetaConditions once we re-derive the decorrelation for 2017 and 2018

simonepigazzini:
Which syst are missing? We know from ttH that electron Reco and ID are not dumped correctly, anything else? (ele SF should be fixed by a PR from the ttH group).

Systematics/test/data_2017.json
edjtscott: Yes good idea, will do that, thanks

MicroAOD/python/flashggJets_cfi.py
edjtscott: This is just moving the definition, which was previously [here](https://github.com/cms-analysis/flashgg/pull/1169/files/e931a2cb0000855f95acb164c9ec402267499d71#diff-209dbe06f63ca00a7b125b7fea7269feL10-L26), all functionality unchanged

edjtscott:
> Which syst are missing? We know from ttH that electron Reco and ID are not dumped correctly, anything else? (ele SF should be fixed by a PR from the ttH group).

We thought some other systematics were missing too (JECs, MET-related) but actually this is not the case, they are present in the test workspaces. So in principle this could be ready to merge and the incorporation of the other workflows addressed in a later PR

Taggers/interface/CollectionDumper.h
threiten: Is this actually doing the reweighting (i.e. multiplying the NNLOPS weights with weights) or just dumping the NNLOPS weights ?

Taggers/interface/CollectionDumper.h
```diff
@@ -31,7 +31,9 @@
 
 #include "flashgg/MicroAOD/interface/CutBasedClassifier.h"
 #include "flashgg/MicroAOD/interface/ClassNameClassifier.h"
+#include "flashgg/MicroAOD/interface/StageOneNameClassifier.h"
```
simonepigazzini, line 34: Looking more into the details I am wondering if we need the HTXS splitting to be hardcoded into the CollectionDumper. Why can't we use the CutBasedDumper using a cut on the HTXS bin number to define the categories?

Taggers/interface/CollectionDumper.h
edjtscott: It actually applies the reweighting; this is already present in the legacy branch however, see [here](https://github.com/cms-analysis/flashgg/blob/dev_legacy_runII/Taggers/interface/CollectionDumper.h#L615-L621)

Taggers/interface/CollectionDumper.h
edjtscott: it's a good point, and I can't recall the reason Seth preferred to do it this way originally. Perhaps we can discuss in person whether it's worth restructuring this or not.

Taggers/interface/CollectionDumper.h
threiten: Ok thanks. The version of the CollectionDumper is a little different there in the differentials branch [(see here)](https://github.com/threiten/flashgg/blob/a95a194d48fbeb271f7d49d3e8c6889d359459df/Taggers/interface/CollectionDumper.h#L638-L653), where it is in fact just dumped. In the version that is in this PR now, the NNLOPSweight will only be applied of splitPdfByStage0Bin_ || splitPdfByStage1Bin_ and reweighGGHforNNLOPS_ are true, which is not true for us. So we might want to find a solution that will apply the NNLOPSweight also if only reweighGGHforNNLOPS_ is true.

simonepigazzini:
@sam, @mhl0116 I am mergin this major update of the code which will likely imply some action on your side to fix conflicts (just warning you here).

@chernyavskaya I suggest you take a look at this update when you have time to see that nothing has been broken on the HH side, although for the time being I suggest you keep running ntuples with PR1186. We might merge once you are ready to produce the final (approval-ready) results.

PR #1170, opened 2019-11-19
fix how lepton related systematics are dumped

mhl0116:
As discussed in #1169, fix how lepton systematics are dumped.


simonepigazzini:
Thank you. merging now

PR #1171, opened 2019-11-21
Fix THQ likelihood discrimitor class

fabio-mon:
- Avoid costheta>1 in THQTag (probably due to roundoff).
- Load all histograms just once at the beginning of the job instead of at every event.
- Rename the LikelihoodClass and move it to Taggers instead of DataFormats


simonepigazzini:
@gourangakole just letting you know that we are fixing this.

simonepigazzini:
Thank you Fabio, I'll merge the PR tomorrow.

simone

PR #1177, opened 2019-12-05
removing hard coded jet eta cut 

panwarlsweet:
-- adding to read jet max eta cut from metacondition json file to read it properly for all three years.


panwarlsweet:
@simonepigazzini  I already notified this to @chernyavskaya 
Could you please merge it?
Thanks!

PR #1178, opened 2019-12-05
HH updates towards preapp

nadya-chernyavskaya:
Updates:
- b-jet energy regression JER
- Dalitz decays are now kept at gen level for HH analysis


simonepigazzini:
This PR supersedes PR #1167

PR #1181, opened 2019-12-09
Updated ttH MVAs

sam-may:
Hi Simone,

This PR updates the ttH Hadronic and Leptonic taggers to use the latest DNNs and BDTs that were used for the ttH signal strength + CP paper approval.

In addition to updating the weight files, there are some substantial changes to the code because of the fact that we added a preprocessing scheme to the input features for the DNNs.

The relevant weight files are located here (though it looks like Meng already copied these to `eos`):
http://uaf-8.t2.ucsd.edu/~sjmay/ttH/forPR_topic_8Oct2019MVAs/


simonepigazzini:
Hi @Sam. Sorry for not spotting this before ... but could you please change the naming convention such that it is clear that all DNN related files (DNNHelper, DNNPreprocessingConstructor and so on) are clearly identified as part of the TTH workflow? something like TTHDNNHelper would go. This is because, as far as I understand, the classes are not generic but only work for the DNN architecture used for the TTH tagging.

Thank you.

PS: please also add at the bottom of every cpp file (.h, .cc, .cpp) the comment lines that configure emacs and vi to enforce a consistent indentation (you can find the lines in any cpp file in the repo).

sam-may:
Hi Simone, you're correct -- the DNN class is ttH-specific. The files are renamed more appropriately and relevant cpp files have the text editor config lines at the bottom now.

simonepigazzini:
Hi I am testing a different way of dealing with RAM intensive NN and BDTs. The devs I am making are on top of this PR so I will marge it later. 

@sam-may can you please confirm that the UseLargeMVAs flags is no longer useful? The way you eventually dealt with the memory issue is simply to have a common instance of the TTH taggers for all syst variation right?

sam-may:
@simonepigazzini , yes, when producing workspaces for ttH tags only we used a common instance of the TTH taggers. The `UseLargeMVAs` flag was put in so that it was still possible to run all tags without the TTH tags taking up too much memory.

edjtscott:
hi @sam-may and @simonepigazzini - the developments from this PR are now included in [PR 1204](https://github.com/cms-analysis/flashgg/pull/1204) so I will close this one.

Issue #1174, opened 2019-12-02
Category dumper cleanup


simonepigazzini:
In the attempt to merge the STXS and differential PR (#1169 and #1166) these issues arose:

1) Do we really need to have the STXS splitting hardcoded inside the CategoryDumper, is
    there any show stopper to implement it using a CutBasedDumper instead?

2) Pdf and NNLOPS weights are also hardcoded inside the dumper class. Is this unavoidable or can       
    we write a producer that adds these weights to the TagTruth object and dump them from there?

simonepigazzini:
@threiten

simonepigazzini:
I think for the discussion we could profit from having a look at an example file from the STXS

edjtscott:
Thanks for flagging these two issues, I put an example output workspace from workspaceStd with doStageOne switched on here [1].

The main difference from the "usual" workflow is that the STXS bin value is also dumped, and this is then used later on to split the outputs into files for each STXS process. The dumping of the categories allows one to dump categories named according to the target bin, rather than running a separate producer for each one. As far as I am aware this is the only substantive difference. 

Regarding the weights, I am sure the NNLOPS weights can (and should) be moved. The pdf weights I am not so sure, I am not expert and would have to look through the code in detail. 

Will discuss more tomorrow. 

[1] /eos/user/e/escott/HggLegacy/WorkspaceTest/ForPullRequestReview

edjtscott:
The NNLOPS weights have now been fully factorised from here. I had a quick look and it's not clear that the STXS info in the pdf code can be removed, and anyway the current implementation has been merged into the default branch. I think we can close this issue - there is still something to be worked out regarding how the pdfs work with the useParentDataset info, but this is separate.

simonepigazzini:
I Agree, this is address in #1188 .

PR #1188, opened 2020-01-13
Add workflow for STXS analysis

edjtscott:
This PR introduces all the modifications needed to run the STXS stage 1 workflow, and supersedes PR 1169. 

Additional changes relative to that PR include:
- adding stage 1.1 bin information, which requires moving to CMSSW 10_6_X
- add option for multi-class dijet BDT
- more sensible structure, removing need for additional StageOneTag class
- moving NNLOPS weighting to the TagSorter
- adding a check that the das query works when accessing parent or secondary datasets

The category dumper [issue here](https://github.com/cms-analysis/flashgg/issues/1174) is unchanged - as far as I can tell the code there is necessary for inserting the stage 0 or 1 stxs bins into pdf datasets, so have left it in for now. 

The main unsolved problem is how to incorporate this "standard" workflow with the new ttH-style workflow - to be investigated properly as soon as possible.


DataFormats/interface/DiPhotonTagBase.h
```diff
@@ -34,13 +42,20 @@ namespace flashgg {
...
+        float getTheoryWeight( std::string key ) const;
         void setSystLabel( const std::string label ) { systLabel_ = label; }
         std::string systLabel() const { return systLabel_; }
         bool hasSyst( const string &label ) const { return ( systLabel_ == label );}
         void setIsGold ( int runNumber );
         void setIsGoldMC( bool isGold ) { isGold_ = isGold; }
         bool isGold() const { return isGold_; }
         virtual DiPhotonTagBase::tag_t tagEnum() const { return DiPhotonTagBase::kUndefined; }
+
+        int stage1recoEnum() const { return stage1recoTag_; }
```
simonepigazzini: maybe rename to stage1recoTag to be consistent with the set method defined on line 57

MetaData/python/JobConfig.py
```diff
@@ -5,6 +5,12 @@
 
 import commands
 
+def safe_das_query( search, cmd ):
```
simonepigazzini, line 8: thank you for this

MicroAOD/interface/PhotonIdUtils.h
```diff
@@ -142,8 +143,8 @@ namespace flashgg {
             /// 
             /// }
             unsigned short nSaturated = 0, nLeRecovered = 0, nNeighRecovered = 0, nGain1 = 0, nGain6 = 0, nWeired = 0;
-            auto matrix5x5 = lazyTool.matrixDetId(seed,-2,+2,-2,+2);
-            for(auto & deId : matrix5x5 ) {
+            auto matrix5x5 = CaloRectangleRange(2, seed, *topology);
```
simonepigazzini, line 146: Is this the implementation I made and shared with @jonathon-langford ?

MicroAOD/python/flashggJets_cfi.py
```diff
@@ -251,5 +251,11 @@ def addFlashggPuppiJets(process,
                                   inputTagJets= PuppiJetCollectionVInputTag
 )
 
-  
+flashggUnpackedJets = cms.EDProducer("FlashggVectorVectorJetUnpacker",
```
simonepigazzini, line 254: why is this here now? I think the unpacking is done while running on microAODs not during the microAOD production.

README.org
```diff
@@ -3,14 +3,17 @@
...
 
    Supported releases:
    - 10_5_0
+   - 10_6_1_patch2 required for the STXS stage 1.1 information
 
    get flashgg:
    #+BEGIN_EXAMPLE
-   cmsrel CMSSW_10_5_0
-   cd CMSSW_10_5_0/src
+   export SCRAM_ARCH=slc7_amd64_gcc700
```
simonepigazzini, line 14: this usually goes automatically, can you please remove the SCRAM_ARCH setting from the readme? It's better to keep it simple.

Systematics/python/SystematicDumperDefaultVariables.py
```diff
@@ -9,7 +9,7 @@
 
 minimalNonSignalVariables = ["CMS_hgg_mass[160,100,180]:=diPhoton().mass"]#,"centralObjectWeight[1,-999999.,999999.] := centralWeight"]
 
-minimalVariablesHTXS = minimalVariables+["stage0cat[72,9.5,81.5] := tagTruth().HTXSstage0cat"]
+minimalVariablesHTXS = minimalVariables+["stage0bin[72,9.5,81.5] := tagTruth().HTXSstage0bin"]
```
simonepigazzini, line 12: stage0 ??

Systematics/test/runSTXS/run_data_2016.sh
```diff
@@ -0,0 +1,2 @@
+export X509_USER_PROXY=~/x509up_u72495
```
simonepigazzini: not very user agnostic ;). If you want to keep this scripts in the repo you should change this line such that the script takes the proxy information using 'voms-proxy-info' 

(some for the other scripts)

Systematics/test/workspaceStd.py
```diff
@@ -410,6 +429,8 @@
 # Dump an object called NoTag for untagged events in order to track QCD weights
 # Will be broken if it's done for non-central values, so turn this on only for the non-syst tag sorter
 process.flashggTagSorter.CreateNoTag = True # MUST be after tag sequence cloning
+process.flashggTagSorter.isGluonFusion = cms.bool(bool(customize.processId.count("ggh")))
+process.flashggTagSorter.applyNNLOPSweight = cms.bool(bool(customize.processId.count("ggh")))
```
simonepigazzini: I can't think of a particular use case but maybe we can consider adding an explicit boolean option to activate the NNLOPS reweighting, to be put in AND with the production mode check. What are your thoughts about this?

DataFormats/interface/DiPhotonTagBase.h
edjtscott: good point, will change

MicroAOD/interface/PhotonIdUtils.h
edjtscott: yep it is

MicroAOD/python/flashggJets_cfi.py
edjtscott: good question... I didn't really like it in flashggTags, which already imports things from this MicroAOD config file anyway, see [here](https://github.com/cms-analysis/flashgg/blob/dev_legacy_runII/Taggers/python/flashggTags_cff.py#L3). I am happy to move it to its own config file in Taggers/ if preferred

README.org
edjtscott: we found that this is useful because (at least for some people) the default is now gcc820, and if you follow the steps with that then it won't compile, and it's not at all obvious what is going wrong & how to solve it.

Systematics/test/runSTXS/run_data_2016.sh
edjtscott: aha not very, removed

Systematics/python/SystematicDumperDefaultVariables.py
edjtscott: these are the defaults used if doHTXS is switched on but doStageOne is not. Wanted to keep them because e.g. for the ttH analysis I believe this is the configuration they used. Can remove if preferred of course.

MicroAOD/python/flashggJets_cfi.py
simonepigazzini: I agree it is best to have it here, can you please check that this is changed consistently everywhere?

README.org
simonepigazzini: ok

Systematics/python/SystematicDumperDefaultVariables.py
simonepigazzini: ok

Systematics/test/workspaceStd.py
simonepigazzini: I mean: add an applyNNLOPSWeights to the set of cmd line options, such that the user can switch the reweighting of even for ggH samples.

Taggers/test/stxs_dumper.py
```diff
@@ -5,6 +5,7 @@
 import FWCore.ParameterSet.VarParsing as VarParsing
```
simonepigazzini, line 5: I think it's fine to have this test dumper here under Tagger/test

Taggers/test/runSTXS/run_bkg_2016.sh
```diff
@@ -0,0 +1,2 @@
+
```
simonepigazzini, line 1: if I understand correctly this scripts are used to run the test stxsDumper and not the main results. If so it is fine to have them here otherwise I would prefer to move them to Systematics/test where we are used to keep the main workflow.

Beside the previous note please consider removing the .sh scripts, leaving only the config jsons. An example fggRunJobs command can be added to a Readme file descrbing the STXS workflow or even the github wiki.

MicroAOD/python/flashggJets_cfi.py
edjtscott: yes will do

Systematics/test/workspaceStd.py
edjtscott: got it, will do

edjtscott:
hi @simonepigazzini this should now have everything we discussed implemented. I will do some final checks tomorrow with @chernyavskaya and @mhl0116 @sam-may to ensure this doesn't break their workflows. After that we are hopefully ready to merge.

simonepigazzini:
I apologize for the spam, I merged the wrong PR. The correct one is this one but the same comment I made applies here for you guys:

@sam, @mhl0116 I am mergin this major update of the code which will likely imply some action on your side to fix conflicts (just warning you here).

@chernyavskaya I suggest you take a look at this update when you have time to see that nothing has been broken on the HH side, although for the time being I suggest you keep running ntuples with PR1186. We might merge once you are ready to produce the final (approval-ready) results.

PR #1189, opened 2020-01-13
fixed the crash of THQLeptonicTag for tHq sample

prafullasaha:


simonepigazzini:
@prafullasaha there is way more than just the bugfix in this PR and a lot of conflicts, please make a clean PR with the bugfix only starting from an up to date dev_legacy_runII branch.

PR #1204, opened 2020-03-05
Updated STXS workflow

edjtscott:
The main development of this PR is to update to STXS stage 1.2 and allow both the "usual" workflow and the ttH tag to be run together at the same time with systematics included. Inside the stage one customize function, the usual parallel systematics is used for all tags other than ttH, which have the modified systematics workflow already implemented. 

This also merges the latest developments of the ttH, tHq, and VH leptonic analysts, as well as the latest configurations for the ggH and qqH STXS categories. The ggH multi-class is *not* included, since its performance is not yet understood. 

There is still one remaining issue which should be solved before merging, related to the NNLOPS weights for ggH and having it propagated correctly to the systematic variations. The fix for that should come in the next few days. 

A separate PR will be made for the more granular JEC uncertainties, which I am also investigating now.


edjtscott:
I think this is now ready to merge - @simonepigazzini or anyone else please let me know if you have comments

simonepigazzini (changes requested):
Check data file paths

edjtscott:
Paths should be correct now. Also have the option of the JEC uncertainty splitting added

edjtscott:
Together with updates to the ggH, qqH, and ttH tags

simonepigazzini (approved):

PR #1206, opened 2020-03-23
Fix fatal ROOT error on lxbatch nodes in CMSSW_10_6_X

simonepigazzini:
The following error appeared after moving to CMSSW_10_6_1 while running on lxbatch.

[a] Fatal Root Error: @SUB=TSystem::ExpandFileName
input: $HOME/.root.mimes, output: $HOME/.root.mimes

The root cause is probably the use of some interactive-like features of ROOT.

The fix make sure that the local user environment is copied to the worker node by using the `getenv`
option available in condor.

### DIFF
--- a/MetaData/python/parallel.py
+++ b/MetaData/python/parallel.py
@@ -431,6 +431,7 @@ class HTCondorJob(object):
             fout.write('+JobFlavour   = "'+self.htcondorQueue+'"\n\n')
             fout.write('+OnExitHold   = ExitStatus != 0 \n\n')
             fout.write('periodic_release =  (NumJobStarts < 4) && ((CurrentTime - EnteredCurrentStatus) > 60) \n\n')
+            fout.write('getenv        = True \n')
             fout.write('input         = %s/.dasmaps/das_maps_dbs_prod.js \n' % os.environ['HOME'])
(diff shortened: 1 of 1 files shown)


simonepigazzini:
@mgouzevi I tested the fix using the setup you reported, please double check that everything is in order. If you give positive feedback I will merge the PR.

Best,

simone

@ferriff @panwarlsweet

mgouzevi:
Hi,
thx much. It would be good to profit from this PR to fix also the DEBUG flag in STXS analysis

https://github.com/cms-analysis/flashgg/blob/dev_legacy_runII/Taggers/python/flashggTagSorter_cfi.py#L41

It shall be moved from true to false.

Max

simonepigazzini:
Hi Maxime, the defaults has already been changed in the STXS PR that is currently open, so it will go in with that.

simone

panwarlsweet:
> @mgouzevi I tested the fix using the setup you reported, please double check that everything is in order. If you give positive feedback I will merge the PR.
> 
> Best,
> 
> simone
> 
> @ferriff @panwarlsweet

@simonepigazzini I tested this change. It worked fine for me. Thanks a lot!

PR #1208, opened 2020-04-26
Updated FNUF uncertainties based on the method used in HIG-19-004

rchatter:
The FNUF uncertainties in the default Run-2 branch are the outdated ones from Run-1.
Updated the same in the same Eta-R9 bins based on the derivation by N. Schroeder for
HIG-19-004 [@edjtscott  @simonepigazzini ]


simonepigazzini:
Thanks for the update, merging now

PR #1211, opened 2020-05-11
Update on HH(bbgg), incl. VBF HH

nadya-chernyavskaya:
HH(bbgg) analysis :  
   - VBF HH analysis workflow is now added
   - HH(bbgg) analysis MVA training weights & categorization updated to what is used for the unblinding
   - Cleaning up of redundant functions (e.g. dumpVariablesForData)
   - Adding lepton collections from ttH analysis

Related to all analyses : 
   - fix for QGL that sometimes was nan (now if nan : set to -1)
   - option to save genweight (LHE weight)
   - added fix to ignore photons with negative R9 values (happens very rarely) https://github.com/cms-analysis/flashgg/commit/687bdb4e5df3448920fc84db3a8df15839a7208e


simonepigazzini:
@edjtscott I don't think this will affect in any ways STXS, except for the negative R9 fix that was giving issue with one ttH sample.

@fabio-mon you might want to check this out and make a PR in case you have anything to add on top of this.

edjtscott:
@simonepigazzini @chernyavskaya thanks, all looks good to me, don't foresee any issues with the STXS workflow

PR #1213, opened 2020-05-25
Update diphoton vertex SF/uncertainties for full runII

youyingli:


Systematics/test/workspaceStd.py
```diff
@@ -408,8 +408,9 @@
             variablesToUse.append("PreselSF%s01sigma[1,-999999.,999999.] := weight(\"PreselSF%s01sigma\")" % (direction,direction))
             variablesToUse.append("electronVetoSF%s01sigma[1,-999999.,999999.] := weight(\"electronVetoSF%s01sigma\")" % (direction,direction))
             variablesToUse.append("TriggerWeight%s01sigma[1,-999999.,999999.] := weight(\"TriggerWeight%s01sigma\")" % (direction,direction))
-            variablesToUse.append("FracRVWeight%s01sigma[1,-999999.,999999.] := weight(\"FracRVWeight%s01sigma\")" % (direction,direction))
-            variablesToUse.append("FracRVNvtxWeight%s01sigma[1,-999999.,999999.] := weight(\"FracRVNvtxWeight%s01sigma\")" % (direction,direction))
+            if (customize.processId.count('ggh') or customize.processId.count('GluGluHToGG')) and not customize.processId.count('gghh'):
```
simonepigazzini: @edjtscott @jonathon-langford would it work to have weights added only for same samples and not for other down the line (i.e. inside the mighty finalFits) ?

edjtscott: thanks for spotting this - it's probably better to keep the weights for all processes. The `weight()` function doesn't crash if the weight doesn't exist anyway, just returns 1, so the `if` statement here can be removed. I will have a slight update to how these kind of weights are treated in my next PR anyway, which will arrive soon (~one week)

simonepigazzini: @youyingli can you please remove the if statement and push the change?

youyingli: @simonepigazzini it's done.

simonepigazzini: Thanks

PR #1218, opened 2020-06-16
Updated STXS analysis (work in progress)

edjtscott:
This is the latest version of the STXS analysis, with updated VH MET and tHq tags. There are also minor changes to the weight handling, and a source of uncertainty for the HEM issue is added. 

This branch has not been fully tested, and the ggH multiclassifier is not included yet. So please do not merge until further notice.

PR #1219, opened 2020-06-16
2 categories VBFHH training

nadya-chernyavskaya:
- VBFHH with 2 categories in sync with current flashgg master. "Green" strategy for VBFHH as agreed on 18.06.2020 at Hgg working meeting. BDT is trained using a mix of SM and C2V=0 signals in 2 MX regions </>500 GeV. 1 MVA category is optimized in each MX region.

 - Pdf systematics are now propagated if the PdfWeight is ON. Before it was ON even if PdfWeights are off, but doSystematics are on, which of course did not make any sense. 

 - HEM systematic for 2018 is cherry-picked from the STXS PR #1218. I can revert it back if requested, for now I put it in, to share the latest final PR for ttH+HH combination

### DIFF
--- a/Systematics/test/workspaceStd.py
+++ b/Systematics/test/workspaceStd.py
@@ -335,7 +335,7 @@ if is_signal:
             if customize.doGranularJEC:
                 for sourceName in customize.metaConditions['flashggJetSystematics']['listOfSources']:
                     jetsystlabels.append("JEC%s%s01sigma" % (str(sourceName),direction))
-            if customize.metaConditions['flashggJetSystematics']['doHEMuncertianty']:
+            if customize.metaConditions['flashggJetSystematics']['doHEMuncertainty']:
                 jetsystlabels.append("JetHEM%s01sigma" % direction)
             metsystlabels.append("metJecUncertainty%s01sigma" % direction)
             metsystlabels.append("metJerUncertainty%s01sigma" % direction)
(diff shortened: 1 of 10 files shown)


Systematics/test/workspaceStd.py
```diff
@@ -507,7 +507,7 @@
               currentVariables = []
       isBinnedOnly = (systlabel !=  "")
       is_signal = reduce(lambda y,z: y or z, map(lambda x: customize.processId.count(x), signal_processes))
-      if ( customize.doPdfWeights or customize.doSystematics ) and ( (customize.datasetName() and customize.datasetName().count("HToGG")) or customize.processId.count("h_") or customize.processId.count("vbf_") or is_signal ) and (systlabel ==  "") and not (customize.processId == "th_125" or customize.processId == "bbh_125"):
+      if ( customize.doPdfWeights and customize.doSystematics ) and ( (customize.datasetName() and customize.datasetName().count("HToGG")) or customize.processId.count("h_") or customize.processId.count("vbf_") or is_signal ) and (systlabel ==  "") and not (customize.processId == "th_125" or customize.processId == "bbh_125"):
```
simonepigazzini, line 512: @edjtscott watch out for this. I think the AND makes more sense but might require changes to your workflow (i.e. make sure doPdfWeights is set to true)

edjtscott: thanks for the alert - we should be safe provided the default is left as true

simonepigazzini: perfect thanks

PR #1220, opened 2020-06-25
udpating 2018 MC and Data GT; also fixing JER SF calcuation

panwarlsweet:
- updating 2018 SF for MC and Data where bug related to JER SF has been fixed
slide 6 https://indico.cern.ch/event/920726/contributions/3868370/attachments/2055396/3446379/20-06-11_News_PPD.pdf

- Fixing the JER SF calculation as recommend by JME converens in the HN post [1] to make it pt dependent.
https://hypernews.cern.ch/HyperNews/CMS/get/jes/1006/1/1/1.html

Tested setup with 2018 GTs and new updates in the smearing sf calculation. works fine

For more deatils of pt related SF, this TWiki can be read https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetResolution#Smearing_procedures

### DIFF
--- a/MetaData/data/MetaConditions/Era2018_Prompt_v1.json
+++ b/MetaData/data/MetaConditions/Era2018_Prompt_v1.json
@@ -1,8 +1,8 @@
 {
     "globalTags" :
     {
-        "data" : "102X_dataRun2_Prompt_v15",
-        "MC" : "102X_upgrade2018_realistic_v19"
+        "data" : "102X_dataRun2_Prompt_v16",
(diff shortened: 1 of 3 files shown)


simonepigazzini:
@edjtscott @malcles do we intend to include this updates for all analyses?

edjtscott:
yes, all analyses should update with this. Since it should be a very small effect, we think it's fine if analyses in a rush (e.g. for an ICHEP PAS) don't delay only for this reason though

PR #1221, opened 2020-06-26
Upload of 2017 UL campaign

alesauva:
Upload of 2017 catalogue with some signal, background and data datasets with a simple MetaConditions file that should be updated with the latest 2017 UL elements once available.


simonepigazzini:
@alesauva thanks a lot for taking care of this.

PR #1230, opened 2020-09-29
Fix the bug for high/low weight due to MC pileup overwritten for UL2017

youyingli:
- Because when JobConfig.py includes the pileup information,

   mix_2017_25ns_UltraLegacy_PoissonOOTPU_cfi.py 
   mix_2018_25ns_JuneProjectionFull18_PoissonOOTPU_cfi.py 

   in CMSSW [1] come from the same object which is mix_probFunction_25ns_PoissonOOTPU_cfi.py
   Therefore, if mix_2017_25ns_UltraLegacy_PoissonOOTPU_cfi.py is first included and then 
   mix_2018_25ns_JuneProjectionFull18_PoissonOOTPU_cfi.py, the information from later will overwrites the previous which is 
   UL17 and vice versa. This results in huge weights (2018 MC pileup are mis-assigned to UL2017 samples). Now, these two 
   are pulled to flashgg and modified to independent formats and bugs are fixed.
- Note that UL2017 provides only 99 bins for pileup, so I made UL_runII_v1_2017.json as UL2017 example. 
- Commit for diphoton vertex SF of UL2017 is dummy.

[1]
https://github.com/cms-sw/cmssw/blob/CMSSW_10_6_X/SimGeneral/MixingModule/python/mix_2017_25ns_UltraLegacy_PoissonOOTPU_cfi.py#L2
https://github.com/cms-sw/cmssw/blob/CMSSW_10_6_X/SimGeneral/MixingModule/python/mix_2018_25ns_JuneProjectionFull18_PoissonOOTPU_cfi.py#L2

### DIFF
--- a/MetaData/python/JobConfig.py
+++ b/MetaData/python/JobConfig.py
@@ -181,15 +181,14 @@ class JobConfig(object):
         except Exception:
             print "Failed to load 94X_mc2017 mixing"
 
-	try:
-	    from SimGeneral.MixingModule.mix_2017_25ns_UltraLegacy_PoissonOOTPU_cfi import mix as mix_UL17
-	    self.pu_distribs["Summer19UL17"] = mix_UL17.input.nbPileupEvents
-	except Exception:
-	    print "Failed to load UL17 mixing"
-
+        try:
+            from flashgg.MetaData.mix_2017_25ns_UltraLegacy_PoissonOOTPU_cfi import mix as mix_UL17
+            self.pu_distribs["Summer19UL17"] = mix_UL17.input.nbPileupEvents
+        except Exception:
+            print "Failed to load UL17 mixing"
 
         try:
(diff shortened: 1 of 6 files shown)


youyingli:
Hi @edjtscott and @Joe-W-Davies, could you please check if this bug-fixing PR can solve the high/low weight problem mentioned from

https://indico.cern.ch/event/943905/contributions/4033510/attachments/2111058/3551063/UL_vs_ReReco_2017.pdf

edjtscott:
hi @youyingli yes for sure, we are going to include it in our next ntuple production to check the weights are returned to normal. We will let you know as soon as it's done

edjtscott:
Thanks a lot @youyingli for this PR - @Joe-W-Davies has now tested it and confirmed that the weights now behave sensibly

youyingli:
Thank @edjtscott @Joe-W-Davies for confirmation. The PR is ready to merge.

PR #1233, opened 2020-10-07
Include the ggH multiclass BDT

edjtscott:
This is the final missing piece of the STXS analysis in HIG-19-015. The producer for the GluGluHMVAResult and its implementation with an option (default true) in the main stage one tag producer are included, as well as the small tweaks necessary to the python files. 

I have tested locally but probably best to wait for confirmation from @fabio-mon that the results are sensible.


youyingli (approved):
There is basically no problem to merge to the PR of overall STXS framework. However, to put the slide or reference link here about this new method (ggH MVA for STXS categorization) might be useful.

PR #1236, opened 2020-10-28
Fix stxs dumper

edjtscott:
A couple of small changes to make the existing stxs_dumper compatible with the latest updates to flashgg. Thanks to Ilya for spotting the crashes.


youyingli (approved):
This can be quickly merged when including "TagTruth" because now it's no longer added in STXS.

PR #1253, opened 2020-11-30
Update ttH tags for ultralegacy

sam-may:
Update ttH tags for UL with retrained MVAs, new pT bins, and new SR boundaries.

Summary:
- Add a 5th pT bin for both ttH Hadronic and ttH Leptonic, splitting the [200, inf] bin into [200, 300] and [300, inf]
- Retrain DNNs and BDTs (now including the 2017 UL datasets)
- Rederive signal region boundaries: number of tags per pT bin is 2, except for the [200, 300] and [300, inf] pT bins in the Leptonic channel, which have 1 bin each (due to low number of events in data).

Expected sensitivity for each pT bin is documented here:
https://github.com/sam-may/ttH/blob/master/Binning/results/stxs_results_legacy_v1.1_27Nov2020_cl_v3.json

The relevant weight files have been uploaded to `/eos/cms/store/group/phys_higgs/cmshgg/flashgg-data/Taggers/data/`


youyingli:
Thanks for the UL update of ttH STXS. There is no problem for the part of the changed codes. Hi @edjtscott , please also check if it's expected in STXS framework. Thanks!

edjtscott:
Incorporated into PR 1272, will now close this one

PR #1255, opened 2020-12-10
Adding PU re-weighting 2018 UL

Soumyatifr:


youyingli:
Hi @Soumyatifr , thanks for the contribution. However, using the UL pileup source directly from CMSSW will overwrite ones for other years in flashgg and this causes the inconsistent pileup reweight. As #1230 ,  high/low reweight resulted from in  inconsistent pileup reweight. Therefore, I also make another PR #1256 for updating this part. You can follow up this change. Close this PR.

Soumyatifr:
Thank you veru much.

PR #1258, opened 2020-12-17
Fix broken UL2018 catalogue for EGamma data

alesauva:
Fix broken UL2018 catalogue where "nevents" were missing for EGamma run C and D.


simonepigazzini:
what was exactly fixed here?

alesauva:
There was a problem with the publication for some of the EGamma run C and run D files for which I had to force the importation. In the process the corresponding "events:" lines in the catalogue were not present which was giving errors when producing ntuples. I then simply added the lines with the correponding number of events.

PR #1268, opened 2021-03-08
Fix for dumping systematic variation of theory weights for ultra-legacy samples

threiten:
Only tested with ultra legacy 2017 sample. Compatible with ReReco as well (tested on 2017, given the right settings) 
mc2hessian csv file should be confirmed to be correct here: https://github.com/cms-analysis/flashgg/blob/dd6661a55448c403b46d1155510c67a313cd44a8/MetaData/data/MetaConditions/Era2017_legacy_v1.json#L240

This is only a quick fix and it should be reviewed in the future how this information can be accessed in a more consistent way

### DIFF
--- a/MetaData/data/MetaConditions/Era2016_RR-17Jul2018_v1.json
+++ b/MetaData/data/MetaConditions/Era2016_RR-17Jul2018_v1.json
@@ -213,7 +213,7 @@
 	"eta" : 2.4
     },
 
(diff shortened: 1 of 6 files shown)


threiten:
The PR should now be in a state that produces the correct pdfWeight variations for RR16, UL17, RR18. It was tested for those datasets with the differential analysis. But it should be tested again in the main branch with one of the standard analyses.

youyingli:
PR #1285 PR #1287 had fixed the PDF issue for UL and ReReco. Close this PR. Thanks for this contribution.

PR #1272, opened 2021-07-01
Update taggers and configs with final versions for HIG-19-015

edjtscott:
The various tag re-optimisations performed when the UL 17 dataset was incorporated into HIG-19-015 are included in this PR.  So far I have checked for compilation but not yet tested that it runs - probably best to wait that I can confirm I get sensible looking results before merging.


edjtscott:
Update - have tested on 2016 data and yields seem to be consistent with those in my branch (the one used for the final results) following a VH hadronic fix.

youyingli (approved):
Hi @edjtscott, sorry for the very late reply. This PR is fine for me. Thanks!

PR #1289, opened 2022-07-27
Many changes that should be transparent for all the users. Main reason for this: postVFP catalog for RunIISummer20UL18

emanueledimarco:
The reason for this PR is mainly the latest commit ac78a670ba71344c60d1ce0ca39311bae555bfaf which contains the updated postVFP samples catalog. 

The other commits are for the categories of the anomalous couplings in VBF, and additional files for production of ntuples, that should be transparent to all users (I hope). Otherwise, one can cherry-pick only the latest commit above, if it is cleaner and want to postpone or not merge all the rest.


emanueledimarco:
Closing this long PR, leaving aside analysis changes and only pushing the Summer20UL16 catalog (PR #1290 )

PR #1308, opened 2024-02-21
Catalogs after the new MC and data production done to recover deleted files in Saclay Tier2

emanueledimarco:
Catalogs by @ram1123. Since some files are still being recovered and will be included, a new PR will be done later integrating  them (should be more minimal changes with respect these).

@lfinco @fcouderc @federica1998


emanueledimarco:
Since this PR was not merged, I updated it with:
1. 2018 samples are updated after running the full check step to compute the weights. They include back all the AC samples (removed for some reason after the last updates), and the QCD double e.m. enriched missing
2. all the years have included the SHERPA samples for the width with interference (int-gg/int-qg)

@lfinco @fcouderc @federica1998 @raeubaen for info
@ram1123 this should include everything was pending in this re-production
@youyingli this PR should be merged in the central repository, if you agree

PR #1309, opened 2024-05-03
Catalogs update after disks hardware problems in Rome Tier2 in April 2024

emanueledimarco:
Two commits:
* re-run the check scripts to fix the MC catalogs invalidating the files which are now missing in Rome which were lost because of disks broke in April 2024
* add the sample ggH - qg interference for mH=120 GeV


emanueledimarco:
Closed because it is superseded by #1310 
@youyingli can you merge #1310, so I'll make a new one on top of that with update catalogs with the fixed MC sample catalogs after bad file removal?
@lfinco @fcouderc @bmarzocc

